PR #781, opened 2015-01-07
add event when rendering subnodes

erazor83:
I've been a bit confused while reading the docs about onRefresh event. Which actually does not fire when subnodes will be rendered.

Instead of firing onRefresh i decided to use a new event onRenderNode.

Maybe onRefresh, onRender and onRenderNode should be merged into one event which has an additional parameter which holds the type?

### DIFF
--- a/src/w2sidebar.js
+++ b/src/w2sidebar.js
@@ -8,7 +8,7 @@
 * == NICE TO HAVE ==
 *   - add find() method to find nodes by a specific criteria (I want all nodes for exampe)
 *   - dbl click should be like it is in grid (with timer not HTML dbl click event)
-*   - reorder with grag and drop
+*   - reorder with drag and drop
 *   - node.style is missleading - should be there to apply color for example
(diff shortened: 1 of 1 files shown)


vitmalina:
I looked at your comments and I see that there are some confusion, but I am not sure your changes will clear it up. Let me shed some lights on onRender and onRefresh. 

onRender - happens ones when element is initially rendered and it will completely rebuild HTML of the element and reinsert entire component. It is not efficient if it is called often.

onRefresh - will refresh whole tree or a node and its subnodes. You can call it with a node id and then it will refresh only that nodes and its subnodes, which will be much faster on large trees.

One thing to remember about sidebar is that it must work recursively, which means I can call a top node to refresh and it will trigger all subnodes or I can call a a subnode that will refresh itself and all its subnodes.

Also, I have implemented similar change, see https://github.com/vitmalina/w2ui/commit/772a8e7037a5566bb70f504731c525d809dbc616

PR #783, opened 2015-01-09
Alpha versione of a prject based on django that provide data server-side...

FilippoBenassutiCDM:
This is a simple example of integration of w2ui for client side and Django, python web framework, for server side.
This component can be used as app for Django project to be able to use w2ui with server side elaboration.

The example use a sqlite database but is possible switch to another database that is compatible with Django  (es Postgres or mysql ).

Fix to use the server side library to works with others database types may be required.


vitmalina:
I hope it can be useful for people who use Django.

PR #784, opened 2015-01-09
Fix advanced search for list

AlexKovynev:
Fix advanced search for list type with one selection

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -1369,6 +1369,8 @@
                     var value1   = field1.val();
                     var value2   = field2.val();
                     var svalue   = null;
+                    var text     = null;
+
                     if (['int', 'float', 'money', 'currency', 'percent'].indexOf(search.type) != -1) {
                         var fld1 = field1.data('w2field');
                         var fld2 = field2.data('w2field');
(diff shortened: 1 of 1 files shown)


AlexKovynev:
Unfortunately this is only half the problem. Search happens, stored but not displayed in the search form (for a given "URL"). Value is assigned, albeit on a key, but is assigned, and then erased, can not find what the problem is.

AlexKovynev:
I did it:) Perhaps there is a more elegant solution, but it works.

vitmalina:
Please help me understand what exactly fix does. I see that it relates to enum and list field types and if I understand correctly, instead of only checking the id of the selected item, you added logic to check for its text too. Can you please give me a use case. As I see it, we should only search by id, not by text, and in fact you can have id and text be the same and then it will search for both.

AlexKovynev:
Try "searches" element type "list" and in the options assign a "url" and get json. The problem is visible immediately, you have a static list without ids in demos, and its working, I need to get data from the database.

AlexKovynev:
First commit solves the problem of finding, the second commit solves the problem of display.

PR #787, opened 2015-01-22
more explicit definition to fix conflict with foundation.css

mpinkston:
Would it be ok to add a more explicit reset definition for table stylings? 
foundation.css rules aren't being properly reset.
foundation.css defines tables as:
```
table tr th,
table tr td {
  padding: 0.5625rem 0.625rem;
  font-size: 0.875rem;
  color: #222222;
  text-align: left;
}
```

### DIFF
--- a/dist/w2ui-dark.css
+++ b/dist/w2ui-dark.css
@@ -67,14 +67,17 @@
   margin: 0px;
   padding: 0px;
 }
+.w2ui-reset table {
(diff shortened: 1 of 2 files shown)


vitmalina:
You are doing it in the wrong place. All /dist/* files are generated with grunt. You need to make changes in /src/less folder and then it will be compiled into common css.

PR #788, opened 2015-01-22
Get the default values for the new record

AlexKovynev:
Ability to get the default values for the new record.

### DIFF
--- a/src/w2form.js
+++ b/src/w2form.js
@@ -1163,7 +1163,7 @@
             // after render actions
             this.resize();
             var url = (typeof this.url != 'object' ? this.url : this.url.get);
-            if (url && this.recid != 0 && this.recid != null) {
+            if (url) {
                 this.request();
             } else {
                 this.refresh();


src/w2form.js
```diff
@@ -1163,7 +1163,7 @@
             // after render actions
             this.resize();
             var url = (typeof this.url != 'object' ? this.url : this.url.get);
-            if (url && this.recid != 0 && this.recid != null) {
+            if (url) {
```
vitmalina: "this.recid != 0 && this.recid != null" - means that the record is empty in the form. So, no request should be made. 

As for default values, I think they should be pulled one type somewhere else and filled out for the form. You can easily add to onRequest method and remember them, so that next time no request is made. 

Also, changing it here will entail changing it in .reload() function, which has same condition. And .clear() function, plus if you do want to fill out default values anyway, you can call .request() on your own and it will submit a request.

PR #794, opened 2015-01-25
Synchronize "dist" with "src"

AlexKovynev:
Fix for 793


vitmalina:
I do not recompile into dist folder daily, but a couple times a week. I have checked, your changes made it.

PR #811, opened 2015-02-09
Mind the checked state of check, drop, menu when refreshing the toolbar

mpinkston:
This request applies/removes the 'checked' class for check/drop/menu items when the toolbar is refreshed.

### DIFF
--- a/src/w2toolbar.js
+++ b/src/w2toolbar.js
@@ -362,6 +362,14 @@
                 el.html(html);
                 if (it.hidden) { el.css('display', 'none'); } else { el.css('display', ''); }
                 if (it.disabled) { el.addClass('disabled'); } else { el.removeClass('disabled'); }
+                if (it.type == 'check' || it.type == 'drop' || it.type == 'menu') {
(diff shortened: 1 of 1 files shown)


src/w2toolbar.js
```diff
@@ -362,6 +362,14 @@
                 el.html(html);
                 if (it.hidden) { el.css('display', 'none'); } else { el.css('display', ''); }
                 if (it.disabled) { el.addClass('disabled'); } else { el.removeClass('disabled'); }
+                if (it.type == 'check' || it.type == 'drop' || it.type == 'menu') {
```
vitmalina: checked class is already added in line 443 (getItemHTML method). I do not see why you need it here too.

mpinkston: Ah, You're right.. This only seems to fix my bug because I'm selecting the button outside of $(this.box) scope.
In my case, (this) seems to reference an instance that is somehow detached. 
Here's a fiddle: http://jsfiddle.net/mpinkston/fayjfxhe/

Issue #810, opened 2015-02-08
w2confirm inconsitent default action


gadelat:
When w2confirm is created, "No" button is focused, but ENTER key triggers "Yes" action.
I would create PR for this myself, but I'm not sure which action is intended to be as default.

vitmalina:
Focus does not seem to be visible. I have looked up in the code and I see this for w2confirm
- enter key will trigger Yes answer
- esc key will trigger No

For w2alert both enter and escape key will close it.

For me this sounds like a desired behavior. However, I do agree it might be useful to show visually which button is in focus. And possibly allow developer to choose which button is in focus initially.

gadelat:
You are focusing it here: https://github.com/vitmalina/w2ui/blob/master/src/w2popup.js#L893 Focus on that button is visible in Firefox and it's confusing.

vitmalina:
Confirmed. I am going remove that line on FF for now. As I do not see how to make it cross browser consistent. On different browsers it is different.

gadelat:
I don't think main problem here is that some browsers don't show focus of the button. Problem here is that those which do are doing so on the wrong button. No need to remove it, just change it to a correct button.

vitmalina:
I think this issue is no resolved. Please see this weeks commits.

PR #812, opened 2015-02-10
Fix wrong focus

gadelat:
Proposed solution to fix issue #810


vitmalina:
Also, check out my latest changes. I have added a visual mark for buttons in focus. Also, hitting tab will not loop through the elements inside of popup, not entire HTML page.

PR #843, opened 2015-03-13
grid localize bug fix

srapillo:
This wants to be a contribute to solve grid toolbar localization issue.
To complete the fix you need to change w2utils.js (see my w2utils.js file change proposal)

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -281,26 +281,49 @@
 
     // ====================================================
     // -- Implementation of core functionality
+    
+    // set grid toolbar and buttons strings using the current language
+    w2grid.localize = function() {
+        var p = w2grid.prototype;
+
+        p.msgDelete = w2utils.lang('Are you sure you want to delete selected records?');
+        p.msgNotJSON = w2utils.lang('Returned data is not in valid JSON format.');
+        p.msgAJAXerror = w2utils.lang('AJAX error. See console for more details.');
+        p.msgRefresh = w2utils.lang('Refreshing...');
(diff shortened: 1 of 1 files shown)


vitmalina:
What bug is it actually solving? I looked at the changes and do not see what difference will they make. I have made a checking earlier with toolbar localization fix.

vitmalina:
I have provided a fix. See latest commits.

srapillo:
The problem arises when you change the locale some time after having loaded a grid.
We have a web application supporting at least two languages.
The app starts in italian and the operator can change the current locale at run time switching to french (and viceversa).
With the master version when we set the new locale the grid, (but also the other objects) don't gets the new strings even if we destroy and reload it.
Our patch allowed the right behaviour (and your previous version of w2utils.locale() suggested us the solution). I see in your current version that you have removed also your previous code. I suppose you have a new philosofy about the localization. I'll try your new version.

srapillo:
I tried your new code. It works but there is still a little step to do.
As you can see in the following image:
w2uigrid
The "search" button shows a wrong tooltip. This is because it is built in w2uiGrid by:
 'search'   : { type: 'html',   id: 'w2ui-search',
     html: '<div class="w2ui-icon icon-search-down w2ui-search-down" title="'+ 'Select Search Field' +'" '+
     'onclick="var obj = w2ui[$(this).parents(\'div.w2ui-grid\').attr(\'name\')]; obj.searchShowFields();">
     </div>'
},
and the <div> title is not updated in the refresh function.
Thanks in advance for your attention.

PR #846, opened 2015-03-16
Update w2sidebar.js

kibarda:
expandAll to recursively call expandAll ( not collapseAll)


kibarda:
expandAll to recursively call expandAll ( not collapseAll)

vitmalina:
Good catch. Thanks.

PR #848, opened 2015-03-25
Stop requiring recid for summary

gadelat:
recid doesn't make sense for summary rows


vitmalina:
I do not think it would be correct to do that as recid is user extensively internally. Also, functions such as .get(), .set(), etc will not work without it. However, you can generate some fake recid, it just needs to be unique, when you insert a record and it fixes the problem.

PR #851, opened 2015-03-31
There is a problem for rec_bottom div

majinkai:
if grid url is not empty, then add a record to grid, the `rec_bottom` div will set height to 24px


vitmalina:
I need more information about this issue. If url is not empty, then it supposes remote data source, which means more records can be loaded from the server. And grid.total is the total number of records that can be loaded, but grid.buffered, is the number that is actually loaded. Therefore, it would be more correct to use grid.total in that case.

PR #873, opened 2015-04-26
do not use 'for in' loops on arrays

avistramer:
Emberjs adds methods and properties to the native Array prototype so it can observe/react to changes.  What this means is that any library that does the following:

```
for(var x in someArray) {}
```

will get unexpected results because ```x``` will get values other than just array indices.  I'm hoping to use the excellent w2grid in my project, but the above was causing me problems.  I've gone through the code and replaced these occurrences with simple ```for(var i = 0; i < someArray.length; i++) {}``` style loops.  Hoping this is acceptable so I don't have to maintain a fork.  I suspect Ember is not the only library out there that does something like this, so these changes improve the general compatibility of the grid with other libraries.

Thanks!

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -250,8 +250,8 @@
                 object.records[r] = $.extend(true, {}, records[r]);
             }
             // add searches
-            for (var c in object.columns) {
-                var col = object.columns[c];
+            for (var i = 0; i < object.columns.length; i++) {
+                var col = object.columns[i];
                 if (col.searchable == null || col.searchable === false || object.getSearch(col.field) != null) continue;
                 var stype = col.searchable;
                 var attr  = '';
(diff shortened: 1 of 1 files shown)


vitmalina:
For the last year or so, I avoid using for (var in someThing) if it is an array. Therefore, I agree with your refactoring. But you missed at least several places where loop variable was not replaced with new one. :)

avistramer:
Thanks @vitmalina.  I took another look and found 2 more locations to replace, will send another pull request.  All the other for-in loops appear to be on objects.  I'm sure there are others in other w2ui components, but my focus was on the grid only.

vitmalina:
I found 2 places too. I will check them in, so, please do rebase before submitting.

PR #874, opened 2015-04-28
Removed file attribute from newItem object

mpinkston:
This update resolves an exception that occurs when parametrizing the value for a 'file' field. 
If the File object is contained within the record when it is parameterized by $.param, a 'Uncaught TypeError: Illegal invocation' is thrown. 
By removing this attribute, the form is successfully parameterized and sent to the server.

### DIFF
--- a/src/w2fields.js
+++ b/src/w2fields.js
@@ -2295,8 +2295,7 @@
                 type     : file.type,
                 modified : file.lastModifiedDate,
                 size     : file.size,
-                content  : null,
-                file     : file
+                content  : null
             };
             var size = 0;
             var cnt  = 0;


vitmalina:
I have some questions about this approach. Would it be better if removing of the attribute happened before AJAX request is made and on the data that is sent. You can also do it in the onAdd event handler. The reason I suggest this is that the developer might actually need this value in some use case and it is easy to clean it w/o modifying the lib. Additionally, there is also .content property that gets set later on if FileReader is available, which might break your code.

Thoughts?

mpinkston:
I agree that the original File object should be available to the developer, though my concern is that it breaks out of the box as it is. I don't think the developer should necessarily be required to prune the object before it can be successfully parameterized and sent to the server. Perhaps it could be made available in the before add event? something like: 
                var eventData = obj.trigger({ phase: 'before', type: 'add', target: obj.el, file: newItem, original: file, total: cnt, totalSize: size });
This way the developer can still access the original file object (even add it back to the newItem object), but by default, the field value can be successfully parsed by $.param

The .content property seems to be unaffected by this and works as expected.

(My implementation works around this by rebuilding and sending only the full base64 string, but it was a bit confusing at first when all I was getting was that Illegal Invocation exception)
Other than that, I really like the clean intuitive behavior of this widget.

vitmalina:
agree, will revisit

vitmalina:
It has been fixed for the form. So, out of the box it will work. See https://github.com/vitmalina/w2ui/commit/464bb97e88910e36d0007ec7910bea50561e54ce

PR #877, opened 2015-04-29
replace a few more for-in loops on arrays

avistramer:


vitmalina:
w2grid is by far most involved. Other modules have much less number of loops, plus I might have changed them already.

Issue #879, opened 2015-04-30
mergeChanges two issues


githabas:
Recently I have made some corrections to mergeChanges function. It solves two problems for me:

1) If field name has `-` in his name `eval('record' + key + ' = changes[c][s]')` produces error "ReferenceError: invalid assignment left-hand side".

2) If `field type` is of type `list` or `combo`, merged result was just `[Object]`.

Here is my corrected function. I will be glad if you find corrections useful:

```javascript
mergeChanges: function () {
    var changes = this.getChanges();
    for (var c = 0; c < changes.length; c++) {
        var record = this.get(changes[c].recid);
        for (var s in changes[c]) {
            if (s == 'recid') continue; // do not allow to change recid
            if(typeof changes[c][s] === "object") changes[c][s] = changes[c][s].text;
            try { 
                eval('record[\'' + s + '\'] = changes[c][s]');
            } catch (e) {
                console.log('ERROR: Cannot merge. ', e.message || '', e);
            }
            delete record.changes;
        }
    }
    this.refresh();
},
```

vitmalina:
Appreciate your contribution. I have looked at the diff and I think that your code does not do all the use case. Here is a diff

![screen shot 2015-05-02 at 3 02 40 pm](https://cloud.githubusercontent.com/assets/561995/7442941/6c747b4a-f0dc-11e4-82b8-44db4b5497b3.png)
 
For example, it will not merged nested rules. if records
```
records: [
 name: { fname: 'first', lname: 'last' }
]
```
it will not work with it. 

However, those two use cases that you mentioned need to work too. We need to find a solution that works for all of them. Also, if it the future you can submit pull requests, it will be easier to contribute.

githabas:
Sorry I can't find grid example with this records structure
```
records: [
 name: { fname: 'first', lname: 'last' }
]
```

vitmalina:
I think it should be like this
```
records: [
 { name: { fname: 'first', lname: 'last' } }
]
```

You can create an example and then add columns like
```
columns: [
{ field "name.fname", caption: 'First Name' }
]
```

githabas:
Now it merge nested records too.

PR #884, opened 2015-05-05
mergeChanges update

githabas:
It fixes two issues described in https://github.com/vitmalina/w2ui/issues/879

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -2092,11 +2092,13 @@
                 var record = this.get(changes[c].recid);
                 for (var s in changes[c]) {
                     if (s == 'recid') continue; // do not allow to change recid
-                    var key = "." + s;
-                    if (w2utils.isFloat(s)) key = "['" + s + "']";
-                    if (String(s).indexOf("'") != -1) key = "['" + s.replace(/'/g, "\\'") + "']";
+                    if(typeof changes[c][s] === "object") changes[c][s] = changes[c][s].text;
                     try { 
-                        eval('record' + key + ' = changes[c][s]'); 
+			if(~s.indexOf('.')) {
+				eval("record['" + s.replace(/\./g, "']['") + "'] = changes[c][s]");
+			} else {
+				record[s] = changes[c][s];
+			};
                     } catch (e) {
                         console.log('ERROR: Cannot merge. ', e.message || '', e);
                     }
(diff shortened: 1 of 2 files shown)


src/w2grid.js
```diff
@@ -2092,11 +2092,13 @@
                 var record = this.get(changes[c].recid);
                 for (var s in changes[c]) {
                     if (s == 'recid') continue; // do not allow to change recid
-                    var key = "." + s;
-                    if (w2utils.isFloat(s)) key = "['" + s + "']";
-                    if (String(s).indexOf("'") != -1) key = "['" + s.replace(/'/g, "\\'") + "']";
+                    if(typeof changes[c][s] === "object") changes[c][s] = changes[c][s].text;
```
vitmalina: need space after if and before (. See https://github.com/vitmalina/w2ui/wiki/Contribution-Requirement

src/w2grid.js
```diff
@@ -2092,11 +2092,13 @@
...
                     try { 
-                        eval('record' + key + ' = changes[c][s]'); 
+			if(~s.indexOf('.')) {
+				eval("record['" + s.replace(/\./g, "']['") + "'] = changes[c][s]");
+			} else {
+				record[s] = changes[c][s];
+			};
```
vitmalina: No need for trailing semicolumn

src/w2grid.js
```diff
@@ -2092,11 +2092,13 @@
                 var record = this.get(changes[c].recid);
                 for (var s in changes[c]) {
                     if (s == 'recid') continue; // do not allow to change recid
-                    var key = "." + s;
-                    if (w2utils.isFloat(s)) key = "['" + s + "']";
-                    if (String(s).indexOf("'") != -1) key = "['" + s.replace(/'/g, "\\'") + "']";
+                    if(typeof changes[c][s] === "object") changes[c][s] = changes[c][s].text;
                     try { 
-                        eval('record' + key + ' = changes[c][s]'); 
+			if(~s.indexOf('.')) {
```
vitmalina: replace with if (s.indexOf('.') != -1)

vitmalina:
I appreciate your persistence. As this logic becomes more difficult to follow, can you add some unit tests for this method where all use cases are covered. There are already some unit tests in 

qa/tests/w2grid.js

you can follow similar examples as it is for some other methods.

githabas:
Here is jsfiddle with links to updated mergeChanges:
http://jsfiddle.net/5ncv1rb7/9/
The second grid uses nested records.

vitmalina:
Appreciate it. Seems to work nicely. There was some bug in unit test. I have fixed it.

Issue #813, opened 2015-02-10
IE11 - Grid Inline Edit - text selection


SVO404:
http://w2ui.com/web/demos/#!grid/grid-21
Text can't be selected in IE11, when editing a field (any type).

n3tfr34k:
```javascript
$(this.box).on('selectstart', function () { return false; }); // fixes chrome cursor bug
```
https://github.com/vitmalina/w2ui/blob/master/src/w2grid.js#L3859

https://github.com/vitmalina/w2ui/pull/891

vitmalina:
I have implemented and push change where all these lines are removed. There were in other places too. It seems to be working properly everywhere.

PR #891, opened 2015-05-12
IE inline editing text selection now working.

n3tfr34k:
IE inline editing text selection now working.

https://github.com/vitmalina/w2ui/issues/813


vitmalina:
This line of code was needed for older versions of Chrome when during selection or resize it would show a cursor. I think this is no longer valid, and since Chrome is ever-green browser, I will remove this line completely (and from all other places where it was added). Appreciate your pull request, though.

Issue #920, opened 2015-05-31
1.4.2 multiselect after large selection in chrome42 on windows7 really slow


deployable:
Chrome 42.0.2311.90 on windows7.

Select a number of rows (> 200)
Ctrl-click one to deselect. CPU to 100% for 5-10 seconds. 
Ctrl-click the same record to select. CPU to 100% for 20-30 seconds. 

The copy is happening on the ctrl keydown rather than the combination of ctrl-c or ctrl-x. The ctrl keypress by itself only introduces a small pause of 300-500ms when hitting the ctrl key with a number of records selected.  When combined with the mouse click something goes awry. 

This doesn't occur in Firefox on the PC or in Chrome on a Mac. 

I wasn't able to identify exactly where the slowdown was happening as I was finding it hard to identify a  useful breakpoint to start from as most mousedown events with a key combination make it really hard to trigger properly with a break point interrupting them but changing the copy/cut to not run on just the ctrl/cmd key fixed up whatever is causing the select issue.

deployable:
Profiling during the issue

```
          75.84%	30228.5ms 75.84%	focus	
30228.5ms 75.84%	30228.5ms 75.84%	 jquery.js:4588jQuery.event.special.focus.trigger	
30228.5ms 75.84%	30228.5ms 75.84%	  jquery.js:4258jQuery.event.trigger	
30228.5ms 75.84%	30228.5ms 75.84%	   jquery.js:4894(anonymous function)	
30228.5ms 75.84%	30228.5ms 75.84%	    jquery.js:344jQuery.extend.each	
30228.5ms 75.84%	30228.5ms 75.84%	     jquery.js:137jQuery.fn.jQuery.each	
30228.5ms 75.84%	30228.5ms 75.84%	      jquery.js:4893jQuery.fn.extend.trigger	
30228.5ms 75.84%	30228.5ms 75.84%	       jquery.js:7478jQuery.each.jQuery.fn.(anonymous function)	
30228.5ms 75.84%	30228.5ms 75.84%	        w2ui.js:3846w2grid.keydown	
30228.5ms 75.84%	30228.5ms 75.84%	         w2ui.js:1025keydown	
30228.5ms 75.84%	30228.5ms 75.84%	          jquery.js:4390jQuery.event.dispatch	
5089.3ms  12.77%	5089.3ms  12.77%	select	
5089.3ms  12.77%	5089.3ms  12.77%	 jquery.js:4258jQuery.event.trigger	
[...]

PR #921, opened 2015-05-31
#920 copy causing pause with multiSelect 

deployable:
https://github.com/vitmalina/w2ui/issues/920

I refactored the code to only trigger the copy on ctrl/cmd + c/x and pulled the two keys into the one code path to avoid doing to much restructuring.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -2508,7 +2508,7 @@
                 //case 17: // ctrl key
                 //case 91: // cmd key
                 case 88: // x - cut
-                case 88: // c - copy
+                case 67: // c - copy
                     if (empty) break;
                     if (!event.ctrlKey && !event.metaKey) break;
                     var text = obj.copy();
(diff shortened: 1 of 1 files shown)


vitmalina:
The reason to do copy paste on ctr (or cmd key) not on combination of ctr + c (cmd + c) is because of Safari. In case of safari, it would not do the copy paste. However, it should be much faster in current master branch. I have tested with 3k records and copy paste should take fraction of the seconds. 

Please download from master branch and see if it is any better for you.

deployable:
Oh I see Safari triggers the actual copy before the keydown event fires in the browser =/ Thanks, I'll have a go with master.  

Could possibly check for safari then, so at least the unneeded copies are limited to the one browser?

    isSafari  : (navigator.userAgent.toLowerCase().indexOf("safari") > -1 &&
                     navigator.userAgent.toLowerCase().indexOf('chrome') == -1),

vitmalina:
I think that checking if it is safari sounds like a plan. Can you submit updated pull request?

PR #926, opened 2015-06-04
Bugfix: After date changed we get undefined and NaN results with de_D…

n3tfr34k:
…E.json config and unix timestamp from server.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -6033,7 +6033,8 @@
                     }
                     if (tmp[0] == 'date') {
                         if (typeof tmp[1] == 'undefined' || tmp[1] == '') tmp[1] = w2utils.settings.date_display;
-                        data = '<div style="'+ style +'">' + prefix + w2utils.formatDate(data, tmp[1]) + suffix + '</div>';
+                        date = w2utils.isDate(data, tmp[1], true);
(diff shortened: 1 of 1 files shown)


src/w2grid.js
```diff
@@ -6033,7 +6033,8 @@
                     }
                     if (tmp[0] == 'date') {
                         if (typeof tmp[1] == 'undefined' || tmp[1] == '') tmp[1] = w2utils.settings.date_display;
-                        data = '<div style="'+ style +'">' + prefix + w2utils.formatDate(data, tmp[1]) + suffix + '</div>';
+                        date = w2utils.isDate(data, tmp[1], true);
```
vitmalina: Looks like this will create a global variable date, which is not good.

src/w2grid.js
```diff
@@ -6033,7 +6033,8 @@
...
+                        data = '<div>' + prefix + w2utils.formatDate(date ? date : data, tmp[1]) + suffix + '</div>';
```
vitmalina: I see you remove style="'+ style +'" - why?

vitmalina:
I did some updates how dates are handled, please see if it fixes your problem. If it doesn't, then please send me more info or another pull request.

n3tfr34k:
I will try it out after my one week holiday. Sorry for the date variable in global scope and the removed style part. Copy&paste from our production code into my github fork was not the best idea. In production code we use 1.4.2 and not the current master.

PR #940, opened 2015-06-15
XHTML Compatibility

ceedriic:
I need to use w2ui on an (existing) xhtml-encoded webpage,
so I looked at the changes that would be required.

This this diff, I can already run w2ui grid+popup demos using xhtml encoding.
This diff: 
    1) Close tags properly
    2) remove 'nbsp' entity 
    3) Fix and encode attributes correctly
    4) Add xhtml version of grid/grid2/popup

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -1782,7 +1782,7 @@
         searchShowFields: function () {
             var obj  = this;
             var el   = $('#grid_'+ this.name +'_search_all');
-            var html = '<div class="w2ui-select-field"><table>';
+            var html = '<div class="w2ui-select-field"><table><tbody>';
             for (var s = -1; s < this.searches.length; s++) {
                 var search = this.searches[s];
                 if (s == -1) {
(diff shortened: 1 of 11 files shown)


ceedriic:
Are you interested by this (and further) work with XHTML compatibility?

hansgru:
+1

src/w2grid.js
```diff
@@ -4637,32 +4637,32 @@
...
                         '<div class="w2ui-toolbar-search">'+
-                        '<table cellpadding="0" cellspacing="0"><tr>'+
+                        '<table cellpadding="0" cellspacing="0"><tbody><tr>'+
                         '    <td>'+ this.buttons['search'].html +'</td>'+
                         '    <td>'+
                         '        <input type="text" id="grid_'+ this.name +'_search_all" class="w2ui-search-all" tabindex="-1" '+
                         '            placeholder="'+ this.last.caption +'" value="'+ this.last.search +'"'+
                         '            onfocus="clearTimeout(w2ui[\''+ this.name +'\'].last.kbd_timer);"'+
-                        '            onkeydown="if (event.keyCode == 13 && w2utils.isIE) this.onchange();"'+
+                        '            onkeydown="if (event.keyCode == 13 &amp;&amp; w2utils.isIE) this.onchange();"'+
```
vitmalina: This is not right

src/w2grid.js
```diff
@@ -4637,32 +4637,32 @@
...
                         '            onchange="'+
                         '                var grid = w2ui[\''+ this.name +'\']; '+
                         '                var val = this.value; '+
                         '                var sel = $(this).data(\'selected\');'+
                         '                var fld = $(this).data(\'w2field\'); '+
                         '                if (fld) val = fld.clean(val);'+
-                        '                if (fld && fld.type == \'list\' && sel && typeof sel.id == \'undefined\') {'+
+                        '                if (fld &amp;&amp; fld.type == \'list\' &amp;&amp; sel &amp;&amp; typeof sel.id == \'undefined\') {'+
```
vitmalina: And this

vitmalina:
Frankly, I am not sure there is a pressing need to make XHTML compatibility as people no longer crazy about it :). If you want to add closing tags, tbody, and nowrap="nowrap" I will accept your pull request. However, there were some issue with it (see my comments). Also, I did not know that &nbsp; and &amp; are not compatible with XHTML. Is it? I liked them in &nbsp; and &amp form, as it is more readable.

src/w2grid.js
ceedriic: Why?
&, < and > (as well as " or ' in attributes) need to be escaped.
See: http://stackoverflow.com/questions/730133/invalid-characters-in-xml
Even in pure HTML, &amp;amp is the official way, see result: https://jsfiddle.net/w5ry9c21/

ceedriic:
There are 5 predefined entities that are guaranteed to be in every
XML dialect: http://www.edankert.com/overview/components.predefined-entities.html

Ok, I will split, complete and improve my patch in multiple commits,
with the  non-controversial "closing tags, tbody, and nowrap="nowrap"" first,
and the other commited separately.

PR #946, opened 2015-06-17
Seconds support for w2utils.isTime

loxiw:


vitmalina:
Cool. There is a set of unit tests in /qa folder. It does not cover everything, but isTime is there. Please check if it still works and add unit tests for seconds.

PR #951, opened 2015-06-19
Incorrect float rendering when using '.' as thousand separator

githabas:
If you set "groupSymbol" : "." and "decimalSymbol" : "," float '207147.61' in form will be displayed as '20.714.761,00'.

In an attempt to fix this issue I have implemented numObj.toLocaleString function and updated w2ui's string to number functions.

I think that grid rendering speed now is improved and overall international support now is better and more standard.

There are some cons. Field option 'groupSymbol' now works only as true/false flag. And there is no Safari support:
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString#Browser_compatibility


vitmalina:
Ok. I found the pull request. Changes look good. Thanks.

PR #953, opened 2015-06-21
Unit testing seconds for w2utils.isTime()

loxiw:


vitmalina:
Thanks.

loxiw:
Just to know, should we change the work to do/changelogs?

vitmalina:
I usually put a note on change items in the comments on top of the file. Then, during the release, I will update necessary docs and release notes based on that info. It seems to be easy to maintain.

loxiw:
Ok, won't modify it anymore then. Ty :)

PR #959, opened 2015-06-23
Xhtml Compatibility, Take 2

ceedriic:
Ok, here is a more complete work for xhtml compatibility.
With these changes, everything seems to work well in XHML5 mode, which
can be tested by opening the 3 xhtml files of the last commit in a browser.

I've split the changes in separate, logical patches, so hopefully some (if not all)
of these changes can be merged.


n3tfr34k:
Why someone should using XHTML5 instead of "sloppy" HTML5?

ceedriic:
n3tfr34k:
Because you like to find bugs early?
Because you've existing code that use XML encoding?
Because you've learned that a sloppy tag soup will come back to bite you?
Because you've used tools like http://htmlcleaner.sourceforge.net/, and hated it?
Because when you see `<div><input><button></div>`, you wonder if the button is inside the input?
Because you wonder why you need `find('> table > tbody > tr)` when there is no `<tbody>` in your code?

hansgru:
Converting **html** to **pdf** (not as image or print) is not trivial at all. 
To convert from **xhtml** to **pdf** however (e.g. in order to retain the same stuff in the pdf as on the screen, without using images) is a solved problem: http://itextpdf.com/product/xml_worker, pretty trivial to integrate too. 

Many business applications have this requirement (especially for reports or dashboards)

n3tfr34k:
For converting html to pdf we use wkHTMLtoPDF http://wkhtmltopdf.org/.

@ceedriic you are right in some points, but I will still prefer normal HTML5. I only asked this, because I saw that the support was a lot of work for you ;)

hansgru:
> For converting html to pdf we use wkHTMLtoPDF http://wkhtmltopdf.org/.

w2htmltopdf is platform dependent. It's suitable to generate documentation/user manuals in PDF form, but not really for business reports.

ceedriic:
@n3tfr34k: I totally understand that you would prefer normal HTML5 for your application.

But because w2ui is a library, I think w2ui gets improved and more useful if it is compatible
with both HTML5 and XHTML5 apps. HTML5 and XHTML5 compatibility is not hard after all.

src/w2grid.js
```diff
@@ -4637,32 +4637,32 @@
...
                         '<div class="w2ui-toolbar-search">'+
-                        '<table cellpadding="0" cellspacing="0"><tr>'+
+                        '<table cellpadding="0" cellspacing="0"><tbody><tr>'+
                         '    <td>'+ this.buttons['search'].html +'</td>'+
                         '    <td>'+
                         '        <input type="text" id="grid_'+ this.name +'_search_all" class="w2ui-search-all" tabindex="-1" '+
                         '            placeholder="'+ this.last.caption +'" value="'+ this.last.search +'"'+
                         '            onfocus="clearTimeout(w2ui[\''+ this.name +'\'].last.kbd_timer);"'+
-                        '            onkeydown="if (event.keyCode == 13 && w2utils.isIE) this.onchange();"'+
+                        '            onkeydown="if (event.keyCode == 13 &amp;&amp; w2utils.isIE) this.onchange();"'+
```
vitmalina: Are you sure it is right here? When it is part of HTML, should it be \&amp;?

src/w2grid.js
```diff
@@ -4637,32 +4637,32 @@
...
                         '            onchange="'+
                         '                var grid = w2ui[\''+ this.name +'\']; '+
                         '                var val = this.value; '+
                         '                var sel = $(this).data(\'selected\');'+
                         '                var fld = $(this).data(\'w2field\'); '+
                         '                if (fld) val = fld.clean(val);'+
-                        '                if (fld && fld.type == \'list\' && sel && typeof sel.id == \'undefined\') {'+
+                        '                if (fld &amp;&amp; fld.type == \'list\' &amp;&amp; sel &amp;&amp; typeof sel.id == \'undefined\') {'+
```
vitmalina: Same here

src/w2utils.js
```diff
@@ -1303,7 +1303,7 @@ w2utils.event = {
             $(document).off('click', tmp_hide);
             if (typeof tmp_hide === 'function') tmp_hide();
         }
-        if (obj.length > 0 && (obj[0].tagName == 'BODY' || obj[0].tagName == null)) options.contextMenu = true;
+        if (obj.length > 0 && (obj[0].tagName == null || obj[0].tagName.toUpperCase() == 'BODY')) options.contextMenu = true;
```
vitmalina: Any particular reason to reverse it? :)

src/w2grid.js
ceedriic: Yes:
&, < and > (as well as " or ' in attributes) need to be escaped.
See: http://stackoverflow.com/questions/730133/invalid-characters-in-xml
Even in pure HTML, &amp is the official way, see result: https://jsfiddle.net/w5ry9c21/

src/w2utils.js
ceedriic: Yes, because if tagName is null, then I guess 'tagName.toUpperCase()' will crash.
So the null check must be first.

test/buttons.html
```diff
@@ -23,15 +25,14 @@
...
-        Regular HTML Buttons: <br><br>
-        <input type="button" value="Small"> <input type="button" value="Very, Very, Very Long Button">
+        Regular HTML Buttons: <br/><br/>
+        <input type="button" value="Small"/> <input type="button" value="Very, Very, Very Long Button"/>
     </div>
-
 </body>
-</html>
\ No newline at end of file
+</html>
```
vitmalina: I use this files for testing. There is no need for it to be XHTML compatible.

src/w2grid.js
ceedriic: Same answer :)
https://jsfiddle.net/w5ry9c21/

ceedriic:
Cool, thanks!

PR #961, opened 2015-06-23
Add bower to w2ui instead of w2ui-dist

aadrian:
Add bower to w2ui instead of w2ui-dist.
Let bower manage the dependencies under /libs/ so that they can be ignored from git.
Configure bower to use both 1.x and 2.x versions of JQuery.


hansgru:
Thanks.

vitmalina:
I appreciate the effort, but I am not sure I like how it is going. The only folder needed for distrubution with bower is /dist. I would not mind if there was a way to point bower to download only that folder. I do not know if it is possibly, but would be nice. I personally do not like using bower for my projects as it brings too much junk with libraries. But I recognize the need to a package manager. Maybe some one writes something better. But if it is not possibly, I am in favor to continue to use w2ui-dist.

hansgru:
> Font-Awesome and Code Mirror are not dependencies for w2ui. Only jQuery. It is in the project because I use it in demos. :)

that's why there are 2 types of dependencies in bower: runtime, and development ones:
```json
....
"dependencies": {
    "jquery": "~1.10.0"
 },
....
"devDependencies": {
    "jquery-legacy": "jquery#1.11.3",
    "jquery-modern": "jquery#2",
    "font-awesome": "3.0.2",
    "codemirror": "3.15.0"
}
....
```
Also note that for development (/demo) we can use more versions of jquery (with different naming) and for runtime an "official one"

> The only folder needed for distrubution with bower is /dist. I would not mind if there was a way to point bower to download only that folder. 

That's what should happen with the right ignore patterns (and all bower bugs fixed).  However the readme file and the license file from root would better distributed too.

> I personally do not like using bower for my projects

I don't like it either, but for now, we have no better and usable alternative for the browser (it even has IDE support).

vitmalina:
I have released version 1.4.3. It just has bug fixes for 1.4.x branch. I have checked bower, it is pulling correct version when you install it. I will keep it in w2ui-dist for now. Thanks.

PR #977, opened 2015-07-01
Add an option to choose if nodes are collapsible

Delgan:
Thought it could be useful to have groups which are not collapsible. I just changed 3 lines, I do not know if you might find this option useful.


vitmalina:
I think it will not hurt. Thanks for your contribution.

PR #979, opened 2015-07-02
Arrays support in w2grid.get()

loxiw:
I'm not sure if that's the best way to handle the response when is array and returnIndex is true

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -387,6 +387,20 @@
 
         get: function (recid, returnIndex) {
             // search records
+            if ($.isArray(recid)) {
+                var returnrecs = [];
+                for (var i = 0; i < this.records.length; i++) {
+                    if ($.inArray(this.records[i].recid, recid) != -1) {
+                        if (returnIndex === true) returnrecs.push({'store': 'records', 'index': i}); else returnrecs.push(this.records[i]); 
+                    }
+                }
+                for (var i = 0; i < this.summary.length; i++) {
+                    if ($.inArray(this.summary[i].recid, recid) != -1) {
+                        if (returnIndex === true) returnrecs.push({'store': 'summary', 'index': i}); else returnrecs.push(this.summary[i]);
+                    }
+                }
+                return $.isEmptyObject(returnrecs) ? null : returnrecs;
+            }
             for (var i = 0; i < this.records.length; i++) {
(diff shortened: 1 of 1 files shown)


src/w2grid.js
```diff
@@ -387,6 +387,20 @@
 
         get: function (recid, returnIndex) {
             // search records
+            if ($.isArray(recid)) {
+                var returnrecs = [];
+                for (var i = 0; i < this.records.length; i++) {
+                    if ($.inArray(this.records[i].recid, recid) != -1) {
+                        if (returnIndex === true) returnrecs.push({'store': 'records', 'index': i}); else returnrecs.push(this.records[i]);
```
vitmalina: I think to be consistent, it should just return a number. I know that there is no way to know if this number belongs to records or summary, but I think that it would be in line with existing functionality. Developer will be able to check by looking up a record if this is in records or in summary. Plus, summary are much less frequently used.

src/w2grid.js
```diff
@@ -387,6 +387,20 @@
...
+                    if ($.inArray(this.records[i].recid, recid) != -1) {
+                        if (returnIndex === true) returnrecs.push({'store': 'records', 'index': i}); else returnrecs.push(this.records[i]); 
+                    }
+                }
+                for (var i = 0; i < this.summary.length; i++) {
+                    if ($.inArray(this.summary[i].recid, recid) != -1) {
+                        if (returnIndex === true) returnrecs.push({'store': 'summary', 'index': i}); else returnrecs.push(this.summary[i]);
+                    }
+                }
+                return $.isEmptyObject(returnrecs) ? null : returnrecs;
```
vitmalina: Also, I think it should return [] if no records found. For example, you do
```js
var recs = grid.get([1,2,3]);
if (recs.length != 3) alert('error');
```
Otherwise, you will need 2 conditions, (1) check for null, (2) check if my records is returns.

I will do the changes.

loxiw:
I understand your point in the returnIndex output, I thought about that too... but if the user has to look into records and summarys afterwards, the function has almost no utility XD

vitmalina:
Summary is normally very small a few records at most. In majority of my grids, I do not use summary records. But for those that are used, you will need to add extra logic. Possibly, just return records (not its indexes) and analyze them. I do not think it should be a big problem.

PR #989, opened 2015-07-09
Allow to set empty datetime placeholder

githabas:


src/w2fields.js
```diff
@@ -274,7 +274,7 @@
                     };
                     this.options = $.extend(true, {}, defaults, options);
                     options = this.options; // since object is re-created, need to re-assign
-                    if ($(this.el).attr('placeholder') == null) $(this.el).attr('placeholder', options.placeholder || options.format);
+                    if ($(this.el).attr('placeholder') == null) $(this.el).attr('placeholder', options.placeholder == '' ? '' : options.placeholder ? options.placeholder : options.format);
```
vitmalina: options.placeholder is no longer used. You need to define placeholder in HTML code
```html
<input id="date" placeholder="">
```
and if you define empty one, then no default placeholder will be visible. I verified in current master, all is good.

src/w2form.js
```diff
@@ -1039,7 +1039,7 @@
                     case 'toggle':
                         if (w2utils.isFloat(value)) value = parseFloat(value);
                         $(field.el).prop('checked', (value ? true : false));
-                        this.record[field.name] = (value ? true : false);
+                        this.record[field.name] = (value ? value : false);
```
vitmalina: This was already merged

Issue #719, opened 2014-11-15
IE 11 issue


vitmalina:
The list inline editing does not work in IE11 (http://w2ui.com/web/demos/#!gr.... Combo works fine.

codeart1st:
This can be closed

PR #993, opened 2015-07-14
Fixes for list type field.

codeart1st:
Changed my profile from n3tfr34k to codeart1st ;) More fixes coming.

### DIFF
--- a/src/w2utils.js
+++ b/src/w2utils.js
@@ -1287,7 +1287,7 @@ w2utils.event = {
             maxWidth    : null,              // max width if any
             maxHeight   : null,              // max height if any
             contextMenu : false,             // if true, it will be opened at mouse position
-            style       : 'min-width:100%;', // additional style for main div
+            style       : '',                // additional style for main div
             'class'     : '',                // additional class name for main div
             overlayStyle: '',
             onShow      : null,              // event on show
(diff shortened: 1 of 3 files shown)


n3tfr34k:
Windows git integration ist bad. The full commit message is:

#719 The query select both input fields (helper and original) and try to focus both. But we want focus only the first one.

src/w2utils.js
```diff
@@ -1332,7 +1332,7 @@ w2utils.event = {
             '<div id="w2ui-overlay'+ name +'" style="display: none; left: 0px; top: 0px; '+ options.overlayStyle +'"'+
             '        class="w2ui-reset w2ui-overlay '+ ($(this).parents('.w2ui-popup, .w2ui-overlay-popup').length > 0 ? 'w2ui-overlay-popup' : '') +'">'+
             '    <style></style>'+
-            '    <div style="'+ options.style +'" class="'+ options['class'] +'"></div>'+
+            '    <div style="min-width: 100%; '+ options.style +'" class="'+ options['class'] +'"></div>'+
```
vitmalina: I do not undestand what remove "min-width". This div is inside overlay and should take entire width

codeart1st: This happened in chrome in our setup with list field in a w2ui form. Perhaps it was a conflict with our CSS environment. I will check this again next week.

codeart1st: @vitmalina know I remember me why I added this style property. The real bug is a wrong inline style width in the inner div. The simplest solution was the adding min-width: 100% to this inner div. Look at the pictures below.

Bug

DOM

vitmalina: So, is "min-width" needed?

codeart1st: @vitmalina perhaps we can fix the wrong child width calculation. But using min-width instead of it is not bad in my opinion.

vitmalina: I was fixing some other bug in w2overlay and was testing it. It appears that all is working correctly. I will keep it as is for now. Thanks.

PR #1006, opened 2015-07-28
Animation on fields hide/show in form

githabas:
In some applications some form elements may appear or dissapear dynamically depending on values entered in input fields.  If form elements are hiding or showing suddenly, for user it may be difficult to understand what happened. So it would be really nice to have some animation when fields are hiding/showing. It explains relation logic between various form elements.
In my proposed code I implemented hiding/showing of form elements without refreshing the form, so the actions can be animated.

To make show() and hide() work on multitab forms I have to update get() function too. Hope this does not breaks the library somewhere.

### DIFF
--- a/src/w2form.js
+++ b/src/w2form.js
@@ -206,26 +206,18 @@
         show: function () {
             var affected = 0;
             for (var a = 0; a < arguments.length; a++) {
-                var fld = this.get(arguments[a]);
-                if (fld && fld.hidden) { 
-                    fld.hidden = false;
-                    affected++;
-                }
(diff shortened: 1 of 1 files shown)


vitmalina:
Apologies for the delayed response. I agree with your proposal to introduce animation to show/hide fields and not to invoke refresh of entire form. However, I do not see why there is a need to add another param to get() function. Currently it has 2 params get(field, returnIndex). Though second one is not strictly necessary, it was introduced for consistency reasons because there is one in w2grid, w2toolbar, etc. In grid it is especially necessary for speed issues when there are more then 100k records in the records array. Here, it is kind of useful as you can get index of the field in the fields array and then remove it, for example. However, I do not see reason to introduce another param indexOfPage. I know you use in show/hide functions, but you can also use "[name=fieldname]" selector as each input must have a name. In this way no need for third argument in get(). I would like to keep it this way as it is consistent across the library.

PR #1010, opened 2015-08-01
Fixes for local sort / search

ceedriic:
This makes local sorting and searching work better with non-string cell data.
See individual commits for a more detailed explanation.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -674,6 +674,7 @@
             // process sort
             this.records.sort(function (a, b) {
                 var ret = 0;
+                var defaultToString = {}.toString;
                 for (var i = 0; i < obj.sortData.length; i++) {
                     var fld = obj.sortData[i].field;
                     if (obj.sortData[i].field_) fld = obj.sortData[i].field_;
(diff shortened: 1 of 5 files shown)


src/w2grid.js
```diff
@@ -674,6 +674,7 @@
             // process sort
             this.records.sort(function (a, b) {
                 var ret = 0;
+                var defaultToString = {}.toString;
```
vitmalina: This syntax "{}.toString" is very confusing to me. I am not sure I understand what you are trying to do with it. Can you please explain /or rewrite in more conservative way.

ceedriic: Ok, sorry. If you look at the file I just commited below (test/grid4.xhtml), and run it with an
unmodified version of w2ui, you will see weird behaviour if you type 'undefined' or 'object' or '[' in
the search box:

screen shot 2015-08-10 at 14 04 30
screen shot 2015-08-10 at 14 05 34

This is because of the way variables are converted to String for searching:
 - Missing fields are converted to "undefined",
 - The default toString() method convert objects into the string: "[object Object]"

So basically, when searching:
  1)  we don't want to match Objects that do not redefine the standard toString()
        method, so we need to skip it if obj.toString = {}.toString.
  2) But we DO want to match Objects that have a better toString method, like
       our custom objects or, for example, the Date object.

Anyway, I'm still not happy with the sorting function, I will add a few other patches.

ceedriic:
Ok, I think I'm done here with this serie of patches.

You can test the new grid4.xhtml without these patches, and see the problems:
  1) if you type "undefined" or "[" or "object" in the search box, then it matches garbage.
  2) if you type "0.654" in the search box, the row is not matched (actually the cell is wrong).
  3) If you sort on value column, it is a complete mess, and the sort is unstable: the resulting
sort change if you sort the 'Property' or 'Comment' column before. Now the search is stable.

Does that makes sense?

src/w2grid.js
```diff
@@ -704,6 +685,66 @@
             }
             return time;
         },
+    
+        compareRecords: function (a, b) {
```
vitmalina: I think it would be better to avoid creating methods on the grid levels for internal use (if possibly). For example this method is called compare 2 records for sorting purposes and it returns info expected by array.sort and uses internal vairables of the grid such as grid.sortData. It is unlikely that it would ever make sense for the developer who just consumes grid to call it. Therefore, it should be within localSort function and not accessible for the developer.

src/w2grid.js
```diff
@@ -713,6 +754,7 @@
             }
             var time = (new Date()).getTime();
             var obj = this;
+            var defaultToString = {}.toString;
```
vitmalina: As I mentioned before, this is very confusing syntax for me. I am going to refactor it.

vitmalina:
Thank you for you contribution. I am going o merge this request, but will do changes I mentioned in the comments.

ceedriic:
Perfect, thanks for reviewing+merging!

PR #1020, opened 2015-08-18
Animation on field hide() and show() in form

githabas:
The second parameter in hide() and show() functions specifies show/hide duration.

### DIFF
--- a/src/w2form.js
+++ b/src/w2form.js
@@ -203,30 +203,14 @@
             return false;
         },
 
-        show: function () {
-            var affected = 0;
-            for (var a = 0; a < arguments.length; a++) {
-                var fld = this.get(arguments[a]);
-                if (fld && fld.hidden) { 
-                    fld.hidden = false;
-                    affected++;
-                }
-            }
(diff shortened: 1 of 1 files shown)


vitmalina:
I think your commit changes functionality. Right now, I can do this
```js
form.hide('field1', 'field2')
```
but with your commit it will not work. Is there another way this can be done?

githabas:
To preserve functionality of this functions, hide/show duration have to be hardcoded in library (or maybe config file) or separate setDuration() function have to be set. However I don't think that function call in `form.hide('field1', 'field2')` format is quite necessary in forms. They usually do not have as much fields as tables.

githabas:
Ok here is another idea. We can search function parameters for float type and then if found use it as show/hide duration value. Such solution preserves functionality but it wouldn’t look nice.

vitmalina:
Thinking more about this. I would rather sacrifice duration that make it in a way different from other parts of the library. Thanks for you contribution, but I am closing this pull request.

PR #1022, opened 2015-08-19
Fix #1017 and slightly improve w2popup performance

### LINKED ISSUE #1017: Script executed twice while loading body to w2popup (new in w2ui 1.5.x)
Hello.

I found a very annoying bug using the last nightly 1.5.x version.

When I try to open a popup, the `<script>` tag inside the `body` string is executed twice.
This is problematic, because I often use popup to load external page obtained from an ajax request.

With w2ui 1.4.2, script is executed once: http://jsfiddle.net/5jfdg9e5/
It is executed twice with the last release: http://jsfiddle.net/5jfdg9e5/1/

Delgan:
Because the content of `body` attribute is added twice into the document, this fix just remove every `<script>` tags the first time, so they are executed only once at the end.

Moreover, as parsing html string is quite costly, I used a variable to store the result instead of parsing it twice.

### DIFF
--- a/src/w2popup.js
+++ b/src/w2popup.js
@@ -179,7 +179,7 @@ var w2popup = {};
                           '     width: ' + parseInt(options.width) + 'px; height: ' + parseInt(options.height) + 'px; '+
                               w2utils.cssPrefix('transform', 'scale(0.8)', true) + '"' +
                           '>' + '</div>');
-                msg.append(option_body);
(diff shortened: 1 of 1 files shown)


vitmalina:
I did more testing and there are lots of problems. I get this error "Uncaught TypeError: option_body.clone is not a function". There is a test file in /test/popup.html that needs to work. I have reverted this commit.

PR #1025, opened 2015-08-21
Fix Column Groups with Hidden Columns

ceedriic:


vitmalina:
Your changes look good, however, I need to do more testing. I am merging for now.

ceedriic:
Could you take a look at the XXX line?
It was in the original code, but I see no difference if I remove it.
Could we just remove these two lines?

vitmalina:
I think it is necessary if there is a scroll bar. Then it would have a group extend to the column that has scrollbar. I think it looks a bit better.

PR #1029, opened 2015-08-24
Added lang support for Norwegian / Norway

kesteinbakk:
Not tested everything yet


kesteinbakk:
Somewhere to find the full list of phrases? I see that some are missing and it differs between the other langs which phrases are added...

Issue #1031, opened 2015-08-27
'selected' is not working for 'list'


tynar:
Hello 'selected' feature for 'list' is not working. Similar issue was created in #1008.

vitmalina:
Duplicate issue, see #1008

PR #1034, opened 2015-09-03
#1031 If 'value' is empty from record, then get selected item from 'field.options.selected'.

tynar:


tynar:
#1031 If 'value' is empty from record, then get selected item from 'field.options.selected'.

PR #1040, opened 2015-09-10
Trigger onClose when layout changes

baitur:
When layout changes it will trigger onClose on layout if exists

### DIFF
--- a/src/w2layout.js
+++ b/src/w2layout.js
@@ -163,10 +163,9 @@
                     $(pname).scrollTop(0);
                     panelTop = $(current).position().top;
                 }
-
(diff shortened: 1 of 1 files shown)


vitmalina:
I am not sure onClose is a good event to have. I think I know what are you trying to do is to get an event when an object is replaced. I think that may be it would be better to have onConent even that is triggered each time a content is placed into a panel (thru a content or html methods). Besides, your onClose event is only triggered for objects.

vitmalina:
I have implemented onContent event. There is no need for onClose here. Thanks for the pull request

PR #1053, opened 2015-09-28
Added toolbar scrolling for overlapped buttons.

codeart1st:
I think this can be improved later. But for now a basic implementation of it.

### DIFF
--- a/src/less/src/toolbar.less
+++ b/src/less/src/toolbar.less
@@ -6,12 +6,70 @@
     margin: 0px;
     padding: 2px;
     outline: 0px;
+    position: relative;
     background-color: @toolbar-background-color;
     overflow: hidden !important;
     .user-select;
 
     .disabled {
         .opacity(0.3)
+    }
+
+    &.overflowed {
+        .w2ui-toolbar-scroll-left,
+        .w2ui-toolbar-scroll-right {
+            display: block;
+        }
+
+        .w2ui-toolbar-scroll-wrapper > table {
+            padding: 0 18px;
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


vitmalina:
I pulled your branch and tried to see how it works, but was unable. I tried to create a smaller container and no overflow buttons showed up. Then I tried to resize your toolbar and nothing working. Please provide instructions how to test it.

codeart1st:
scrollbar

I tested it with the default demo sides in the project repo. Also I didn't have updated dist files in my pull request. You need to compile it from my repo before you can test it.

codeart1st:
Ok also it seems that resize function in w2toolbar is only called in a w2layout environment (like in the example folder). Without this environment this solution needs a small hack I think. Something like

```
$(window).resize(function () {

   w2ui['toolbar'].resize();
});
```

vitmalina:
I finally got it working. Visually, it looks great. I like the errors and it seems it would be quick easy to add resize call from grid.resize. However, I have a few suggestions:

1. Once overflow button appears in toolbar does not fit, only right button should appear, not both. If user scrolls right, then left button should appear.

2. I think it would be better when user clicks >> button to scroll toolbar to the next "page", basically entire width of the current toolbar (or to the max it is is less). Currently it scrolls about 10px for each click. 

I am merging it to the current master, but please consider my suggestions.

codeart1st:
After one week of holiday I can answer now. I totally agree with both points of your improvement list. I will work on it later next week ;)

Cheers

vitmalina:
Awesome. Thanks.

PR #1054, opened 2015-10-05
Work In Progress: Make OR/AND local search more flexible.

ceedriic:
This pull request is not for merging at that point, but to get early feedback before doing more work.

In the grid, the search data currently consist of two distinct pieces of information:
  - The array in searchData (http://w2ui.com/web/docs/w2grid.searchData)
  - 'AND' or 'OR' in last.logic.

I see two problems with that:
  - This makes the API and code complicated to read, understand and use.
  - This is not very flexible. for example, when working with a time-serie grid,
    I wanted to search for (T > T1 AND T < T2 AND (OR-based full text search on all columns)).

I would like to improve and simplify the code the following way:
  - Completely remove the global 'AND' and 'OR' logic flag (last.logic).
  - Handle the following two new flags for each structure in the searchData array:
    - 'required': if true, this field indicated that the condition MUST be satisfied (behave like AND)
    - 'not': if true, this field invert the condition (so we can add test like string do not begin with "Input_") .

The diff included in this pull request shows the minimal code I'm using right now to implements the 'required' field, in a backward compatible way.
The test loop go through every condition in searchData, and:
 - if the entry has the 'required' flag set, it must be satisfied.
 - if there are entries without the 'required' flag set, at least _one_ of them must be satisfied.
  
What do you think?
Does it make sense for me to complete the code up to the removal of 'last.logic' and associated code?

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -777,7 +777,10 @@
                 this.total = 0;
                 for (var i = 0; i < this.records.length; i++) {
                     var rec = this.records[i];
-                    var fl  = 0;
+                    var orNb = 0;
+                    var orTot = 0;
+                    var andNb = 0;
+                    var andTot = 0;
                     for (var j = 0; j < this.searchData.length; j++) {
                         var sdata  = this.searchData[j];
                         var search = this.getSearch(sdata.field);
@@ -795,6 +798,7 @@
                                 var val3 = sdata.value[1];
                             }
                         }
+                        var fl = 0;
                         switch (sdata.operator) {
                             case 'is':
                                 if (obj.parseField(rec, search.field) == sdata.value) fl++; // do not hide record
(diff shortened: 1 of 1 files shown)


vitmalina:
I have encountered similar requests from some of the projects I was working on. I agree that flexibility with search is limited. I agree, that you suggesting will make it more flexible, but it still will be limited. For example how would you implement this condition

```
(field1 = 1 AND field2 = 2) OR (field3=1 AND field4=2)
```

Currently, as you rightly noted, you can only do

```
field1 = 1 AND field2 = 2 AND field3=1 AND field4=2
```
or
```
field1 = 1 OR field2 = 2 OR field3=1 OR field4=2
```

I was considering implemented this in the search fields, but never got around to do this. I suggest you can do the following: completed rewrite functions from grid that you need, you can extend  w2obj.grid.prototype way, see http://w2ui.com/web/docs/utils. This way it can be something working for your needs and if this will turn out sometime useful, we can think how to replace current localSearch in the master branch.

PR #1062, opened 2015-10-12
fix #1061

### LINKED ISSUE #1061: no scrollbar in grid
if I init a grid with a small value for limit, so that the height of the initial list is less than the body of the grid, I don't have a scrollbar

erdnax123:


src/w2grid.js
```diff
@@ -2217,7 +2217,11 @@
             // after event
             this.trigger($.extend(eventData, { phase: 'after' }));
             // do not refresh if loading on infinite scroll
-            if (this.last.xhr_offset == 0) this.refresh(); else this.scroll();
+            if (this.last.xhr_offset == 0) this.refresh();
```
vitmalina: sounds good. but please if you use else {} then add {} for if too. I will do it this time.

PR #1064, opened 2015-10-15
fix for removal of first item in a toolbar

giladno:


src/w2toolbar.js
```diff
@@ -143,7 +143,7 @@
                 $(this.box).find('#tb_'+ this.name +'_item_'+ w2utils.escapeId(it.id)).remove();
                 // remove from array
                 var ind = this.get(it.id, true);
-                if (ind != null) this.items.splice(ind, 1);
+                if (ind >= 0) this.items.splice(ind, 1);
```
vitmalina: I think it makes no difference. Condition
```js
if (ind != null) 
```
(note I did not use if (ind !== null) ) will always be true if ind is null or undefined. If ind is 0, the condition will not be true. But for the get function, if it does not find the item, it will return null.

vitmalina:
Please retest, if it works, please close pull request

src/w2toolbar.js
giladno:
```javascript
this.get(it.id, true)
```
This line returns an index number - not an object, where 0 is a valid index - which is being ignored by the next line:
```javascript
if (ind != null)
```

vitmalina: I understand that 0 is a valid index. My point is 
```js
if (0 != null) - returns true
if (0 >= 0) - returns true
```
So, there is no difference what to use. Check in browser console for yourself.

PR #1066, opened 2015-10-17
Display menu icons with menu-radio and menu-check

ceedriic:
screen shot 2015-10-17 at 12 07 56


vitmalina:
Appreciate your contribution. But it looks crowded for me when you use so many icons. Plus I am no longer using image icons (font icons only). However, you can add icons to any menu if you wish, just adding it to caption property. It is not just text, but HTML, so adding icons should be easy.

PR #1078, opened 2015-10-23
Add option withoutMinutes to time field.

MikeZhur:
For time field I need to be able to select only hour, not minutes.
I added option withoutMinutes that if set to true eliminate showing
minutes overlay.

### DIFF
--- a/src/w2fields.js
+++ b/src/w2fields.js
@@ -254,11 +254,12 @@
 
                 case 'time':
                     defaults = {
-                        format      : w2utils.settings.timeFormat,
-                        keyboard    : true,
-                        silent      : true,
-                        start       : '',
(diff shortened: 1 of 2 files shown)


vitmalina:
If you need to select just hours, why don't you just create a drop down with hours only. Or just an input field?

MikeZhur:
I think it would be an usefull option for time field. And I don't think in worth to create another control type for this reason.

vitmalina:
I took a closer look and it looks good. I think, however, it would be better to rename "withoutMinutes" to "noMinutes" as it is shorter. I will do that.

Thanks for your contribution.

vitmalina:
I also added it for datetime field.

PR #1079, opened 2015-10-23
Add blockWeekDays option to date field.

MikeZhur:
To be able to block selection of some days of week (for example, weekends)
I added blockWeekDays options to date field. It's array of week day
numbers. (1 - monday, 2 - tuesday, 3 - wensday, 4 - thursday, 5 - friday,
6 - saturday, 0 - sunday)

### DIFF
--- a/src/w2fields.js
+++ b/src/w2fields.js
@@ -239,13 +239,14 @@
 
                 case 'date':
                     defaults = {
-                        format      : w2utils.settings.dateFormat, // date format
-                        keyboard    : true,
-                        silent      : true,
-                        start       : '',       // string or jquery object
-                        end         : '',       // string or jquery object
-                        blocked     : {},       // { '4/11/2011': 'yes' }
-                        colored     : {}        // { '4/11/2011': 'red:white' }
+                        format       : w2utils.settings.dateFormat, // date format
+                        keyboard     : true,
+                        silent       : true,
+                        start        : '',       // string or jquery object
+                        end          : '',       // string or jquery object
+                        blocked      : {},       // { '4/11/2011': 'yes' }
+                        colored      : {},        // { '4/11/2011': 'red:white' }
+                        blockWeekDays : null       // array of numbers of weekday to block
                     };
                     this.options = $.extend(true, {}, defaults, options);
                     options = this.options; // since object is re-created, need to re-assign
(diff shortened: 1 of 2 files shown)


src/w2fields.js
```diff
@@ -239,13 +239,14 @@
...
-                        blocked     : {},       // { '4/11/2011': 'yes' }
-                        colored     : {}        // { '4/11/2011': 'red:white' }
+                        format       : w2utils.settings.dateFormat, // date format
+                        keyboard     : true,
+                        silent       : true,
+                        start        : '',       // string or jquery object
+                        end          : '',       // string or jquery object
+                        blocked      : {},       // { '4/11/2011': 'yes' }
+                        colored      : {},        // { '4/11/2011': 'red:white' }
+                        blockWeekDays : null       // array of numbers of weekday to block
```
vitmalina: I think it should be empty array by default

src/w2fields.js
```diff
@@ -1863,6 +1864,21 @@
                     }
                     // block predefined dates
                     if (this.options.blocked && $.inArray(str, this.options.blocked) != -1) inRange = false;
+
+                    /*
+                    clockWeekDay - type: array or integers. every element - number of week day.
+                    number of weekday (1 - monday, 2 - tuesday, 3 - wensday, 4 - thursday, 5 - friday, 6 - saturday, 0 - sunday)
+                    for block in calendar (for example, block all sundays so user can't choose sunday in calendar)
+                    */
+                    if (this.options.blockWeekDays != null && this.options.blockWeekDays != undefined
```
vitmalina:
```js
if (this.options.blockWeekDays != null) 
```
is the same as 
```js
if (this.options.blockWeekDays != null && this.options.blockWeekDays != undefined)
```

src/w2fields.js
```diff
@@ -239,13 +239,14 @@
 
                 case 'date':
```
vitmalina: this should also be implemented for "datetime" type

src/w2fields.js
```diff
@@ -1863,6 +1864,21 @@
...
                     // block predefined dates
                     if (this.options.blocked && $.inArray(str, this.options.blocked) != -1) inRange = false;
+
+                    /*
+                    clockWeekDay - type: array or integers. every element - number of week day.
+                    number of weekday (1 - monday, 2 - tuesday, 3 - wensday, 4 - thursday, 5 - friday, 6 - saturday, 0 - sunday)
+                    for block in calendar (for example, block all sundays so user can't choose sunday in calendar)
+                    */
+                    if (this.options.blockWeekDays != null && this.options.blockWeekDays != undefined 
+                        && this.options.blockWeekDays.length != undefined){
```
vitmalina: I think instead of
```js
this.options.blockWeekDays.length != undefined
```
we should use
```js
Array.isArray(this.options.blockWeekDays)
```

src/w2fields.js
easyprojects: Ok. I don't mind. Let in be an empty array. I thought it must work if this option was not specified (for backward compatibility), so I added this checks. But default empty array seems to be ok.

PR #1095, opened 2015-11-02
Fixed add w2ui-input class to w2field

hawkxu:
Additional: add outline-style, outline-width to w2ui-input:focus


src/less/src/common.less
```diff
@@ -79,6 +79,8 @@ input:not([type=button]):not([type=submit]).w2ui-input, textarea.w2ui-input {
     background-color: @input-background-color;
     line-height: normal;
     &:focus {
+        outline-style: solid;
+        outline-width: 1px;
```
vitmalina: Why don't you like regular focus style. I think that shadow kind of style looks good. You are welcome to write a custom css style that you can load on top of regular.

src/w2form.js
```diff
@@ -1032,8 +1032,6 @@
                     case 'email':
                     case 'pass':
                     case 'password':
-                        field.el.value = value;
-                        break;
```
vitmalina: Above fields do not have options, therefore there is no point for it to have options object.

hawkxu: I add the class w2ui-input through the w2field init function, the line 1036 cause the above type field bypass w2field.init function.

src/less/src/common.less
hawkxu: sorry, I used firefox 41.0.2, without the two attributes, the focus field no any outline border appeared, I did not test it with other browser.

src/w2form.js
hawkxu: Please take a look at this: http://plnkr.co/edit/tpl:heAMIR7Yae9GJCXW70nP?p=preview

src/less/src/common.less
vitmalina: Chrome. 

I looked up and it seems that in FF it has nice focus border, not 1px though. Plus this code will break nice border in chrome, which is with a blur.

src/w2form.js
vitmalina: I think it is a wrong place to do that. I have added it in refresh function

vitmalina:
I applied some changes

PR #1132, opened 2015-12-25
w2utils.formatters: Use strict equality operators

yaegashi:
The value checks introduced in 61f073bd also hide cells which contain numeric zero because `0 == ''` is evaluated to true in JavaScript.  Use strict equality operators to avoid the problem.

Test case: http://jsfiddle.net/2ttLd18o/


vitmalina:
Good catch

PR #1135, opened 2016-01-02
TreeGrid: more work with expand/collapse/localSearch/localSort

ceedriic:
Hello, I've integrated the latest w2ui-1.5 with an app that use big trees (10k children+),
and I've found a couple more issues with the expand/collapse/search/sort operations.

This looks like a scary diff, but the first big patch is just some refactoring needed before the real fix.

To fix expand+sort, I needed to add a 'noReset' flag to grid.localSort().
I've set this flag inside grid.add(), but this is an improvement unrelated to the TreeGrid problems.

Let me know if you find anything else funny in this expand/collapse/localSearch/localSort path ;)

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -876,129 +876,8 @@
                 this.total = 0;
                 for (var i = 0; i < this.records.length; i++) {
                     var rec = this.records[i];
-                    var fl  = 0;
-                    for (var j = 0; j < this.searchData.length; j++) {
-                        var sdata  = this.searchData[j];
-                        var search = this.getSearch(sdata.field);
-                        if (sdata  == null) continue;
-                        if (search == null) search = { field: sdata.field, type: sdata.type };
-                        var val1b = obj.parseField(rec, search.field);
-                        var val1 = (val1b !== null && val1b !== undefined &&
-                            (typeof val1b != "object" || val1b.toString != defaultToString)) ?
-                            String(val1b).toLowerCase() : "";  // do not match a bogus string
-                        if (sdata.value != null) {
-                            if (!$.isArray(sdata.value)) {
-                                var val2 = String(sdata.value).toLowerCase();
-                            } else {
-                                var val2 = sdata.value[0];
-                                var val3 = sdata.value[1];
-                            }
-                        }
-                        switch (sdata.operator) {
-                            case 'is':
-                                if (obj.parseField(rec, search.field) == sdata.value) fl++; // do not hide record
-                                // only increment "fl" once -> use "else if"
-                                else if (search.type == 'date') {
-                                    var tmp  = (obj.parseField(rec, search.field + '_') instanceof Date ? obj.parseField(rec, search.field + '_') : obj.parseField(rec, search.field));
-                                    var val1 = w2utils.formatDate(tmp, 'yyyy-mm-dd');
-                                    var val2 = w2utils.formatDate(w2utils.isDate(val2, w2utils.settings.dateFormat, true),  'yyyy-mm-dd');
-                                    if (val1 == val2) fl++;
-                                }
-                                // only increment "fl" once -> use "else if"
-                                else if (search.type == 'time') {
-                                    var tmp  = (obj.parseField(rec, search.field + '_') instanceof Date ? obj.parseField(rec, search.field + '_') : obj.parseField(rec, search.field));
-                                    var val1 = w2utils.formatTime(tmp, 'hh24:mi');
-                                    var val2 = w2utils.formatTime(val2, 'hh24:mi');
-                                    if (val1 == val2) fl++;
-                                }
-                                break;
-                            case 'between':
-                                if (['int', 'float', 'money', 'currency', 'percent'].indexOf(search.type) != -1) {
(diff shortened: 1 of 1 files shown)


src/w2grid.js
```diff
@@ -701,7 +702,8 @@
             // process date fields
             obj.selectionSave();
             obj.prepareData();
-            obj.reset();
+            if (!noReset)
```
vitmalina: if you there are more then one line for if, please use {}

vitmalina: I would try to avoid adding more parameters to sort functions, if possible. What exactly in reset method that messes up sort?

src/w2grid.js
```diff
@@ -3706,10 +3731,15 @@
...
                         stops.indexOf(this.records[end+1].w2ui.parent_recid) >= 0)
                         break;
-                    if (this.records[end+1].w2ui.expanded) this.records[end+1].w2ui.expanded = false;
                     end++;
                 }
                 this.records.splice(start, end - start + 1);
+                this.total -= end - start + 1;
+                var url = (typeof this.url != 'object' ? this.url : this.url.get);
+                if (!url) {
+                    if (this.searchData.length > 0)
```
vitmalina: Please add {} or make if fit one line

src/w2grid.js
```diff
@@ -6981,11 +7020,18 @@
...
 
         prepareData: function () {
+            var obj = this;
+
             // loops thru records and prepares date and time objects
             for (var r = 0; r < this.records.length; r++) {
                 var rec = this.records[r];
-                for (var c = 0; c < this.columns.length; c++) {
-                    var column = this.columns[c];
+                prepareRecord(rec);
```
vitmalina: Does not seem to be aligned right

src/w2grid.js
```diff
@@ -3706,10 +3731,15 @@
...
                         break;
-                    if (this.records[end+1].w2ui.expanded) this.records[end+1].w2ui.expanded = false;
                     end++;
                 }
                 this.records.splice(start, end - start + 1);
+                this.total -= end - start + 1;
+                var url = (typeof this.url != 'object' ? this.url : this.url.get);
+                if (!url) {
+                    if (this.searchData.length > 0)
+                        this.localSearch(true);
```
vitmalina: Please add {} or make if fit one line

src/w2grid.js
```diff
@@ -701,7 +702,8 @@
             // process date fields
             obj.selectionSave();
             obj.prepareData();
-            obj.reset();
+            if (!noReset)
+                obj.reset();
```
vitmalina: I would try to avoid adding more parameters to sort functions, if possible. What exactly in reset method that messes up sort?

Also if "if" statement does not fit one lin, please use {}

vitmalina:
Thank you for your contribution. Please reply to my questions and I will merge request.

src/w2grid.js
ceedriic: The (red) lines 6987 and 6988 above are deleted.
"var rec = .." and prepareRecord() are at the same level.
GIT is showing the diff in a weird way, but i think indentation is correct.

src/w2grid.js
ceedriic: The reason I don't like putting "if (condition) do_something" on a single line is that it makes debugging painful. By putting in on two lines: (1) I can put a breakpoint on the second line and (2) when I run step-by-step, I know if the condition has been taken or not.

My guess is that the "if (condition) do_something" style became common in javascript, (unlike C, C++, Java), because there was no good debugger in javascript until recently...

Ok, I've pushed another commit to add braces everywhere. Is that ok?

src/w2grid.js
ceedriic: Ok, sorry for the bad description. The sort is not really messed up, it's that the user experience is really bad, because obj.reset() scroll the view back to the top. So if you have grid.html with a reversed sort, and a lot of lines like that:

screen shot 2016-01-02 at 22 57 56

When you click to open the node (2) at row 4963, the view scrolls back to the top (row 1) which is painful. I've the same problem without the tree view where I've a websocket that slowly calls add() to add entries to the end of a grid, 100 new records every second up to 50k records. If the add() method calls reset(), the view scrolls back to row 1 every second too, which means you cannot scroll and use the grid while loading... This is fixed by this commit.

ceedriic:
Is that OK now?

ceedriic:
Hi Vitmalina,
Thanks for reviewing and merging that patch.

I've been thinking about that 'noReset' flag added to localSort that you don't like.
Maybe I could try to:
   1) remove that flag from localSort.
   2) remove the reset() call inside localSort() (and maybe just inline the part that's really required).
   3) look at localSort() calls and add an explicit reset() after when it makes sense to scroll back to 0.

That would keep the API clean.
Should I try that?
Cedric

vitmalina:
reset() in localSort is probably needed in general. What I wanted to know
is what exactly does reset do that makes it not needed for tree stcutrues.
Maybe it can be applied only for top records. If the flag is absolutely
needed, then it should stay.

ceedriic:
Vitmalina, is there a way for me to contact you directly (chat, email, etc?)

vitmalina:
Yes. My email is vitmalina@gmail.com, skype vmalinouski. If you want to, we
can setup a skype call. I live in San Francisco.

Vit

PR #1152, opened 2016-02-03
add w2prompt

antitim:
Hi. I added w2prompt() by analogy with prompt(). The Foundation took w2confirm().

```javascript
w2prompt({
  label: 'label',
  value: 'Default value',
  title: 'Title of popup'
  ok_text: 'Add',
  cancel_text: 'No'
}, function(value) {
  w2alert(value);
});
```

### DIFF
--- a/src/w2popup.js
+++ b/src/w2popup.js
@@ -989,3 +989,84 @@ var w2confirm = function (msg, title, callBack) {
         }
     };
 };
+
+var w2prompt = function (options, callBack) {
+    var $ = jQuery;
+    
+    options = {
+        label: options.label ? options.label : '',
+        value: options.value ? options.value : '',
+        title: options.title ? options.title : w2utils.lang('Notification'),
+        width: ($('#w2ui-popup').length > 0 ? 400 : 450),
+        height: ($('#w2ui-popup').length > 0 ? 170 : 220),
+        callBack: callBack ? callBack : null
+    }
+    
+    if ($('#w2ui-popup').length > 0 && w2popup.status != 'closing' && w2popup.get()) {
+        if (options.width > w2popup.get().width) options.width = w2popup.get().width;
+        if (options.height > (w2popup.get().height - 50)) options.height = w2popup.get().height - 50;
+          w2popup.message({
+            width   : options.width,
+            height  : options.height,
+            body    : '<div class="w2ui-centered" style="font-size: 13px;"><div class="w2ui-field"><label>' + options.label + ':</label><div><input id="w2promt" style="width: 100%"></div></div></div>',
+            buttons : '<button id="Ok" class="w2ui-popup-btn w2ui-btn">' + w2utils.lang('Ok') + '</button>',
+            onOpen: function () {
+                $('#w2promt').val(options.value);
+                $('#w2promt').w2field('text');
+                $('#w2ui-popup .w2ui-message .w2ui-btn').on('click.w2confirm', function (event) {
+                    w2popup._promt_value = $('#w2promt').val();
+                    w2popup.message();
+                });
+            },
+            onClose: function () {
+                // needed this because there might be other messages
+                $('#w2ui-popup .w2ui-message .w2ui-btn').off('click.w2confirm');
+                // need to wait for message to slide up
+                setTimeout(function () {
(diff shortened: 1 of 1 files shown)


src/w2popup.js
```diff
@@ -989,3 +989,95 @@ var w2confirm = function (msg, title, callBack) {
         }
     };
 };
+
+var w2prompt = function (options, callBack) {
```
vitmalina: I think to make it similar with w2alert and w2confirm, the arguments should be
```js
w2prompt = function (label, title, callBack)
```
Then, if label is an object - assume it is extended options for the prompt

src/w2popup.js
```diff
@@ -989,3 +989,95 @@ var w2confirm = function (msg, title, callBack) {
...
+    var $ = jQuery;
+    
+    options = {
+        label       : options.label ? options.label : '',
+        value       : options.value ? options.value : '',
+        title       : options.title ? options.title : w2utils.lang('Notification'),
+        ok_text     : options.ok_text ? options.ok_text : w2utils.lang('Ok'),
+        cancel_text : options.cancel_text ? options.cancel_text : 'Cancel',
```
vitmalina: Should be w2utils.lang('Cancel')

src/w2popup.js
```diff
@@ -989,3 +989,95 @@ var w2confirm = function (msg, title, callBack) {
...
+    if ($('#w2ui-popup').length > 0 && w2popup.status != 'closing' && w2popup.get()) {
+        if (options.width > w2popup.get().width) options.width = w2popup.get().width;
+        if (options.height > (w2popup.get().height - 50)) options.height = w2popup.get().height - 50;
+          w2popup.message({
+            width   : options.width,
+            height  : options.height,
+            body    : '<div class="w2ui-centered" style="font-size: 13px;"><label style="margin-right: 10px;">' + options.label + ':</label><input id="w2promt"></div>',
+            buttons : '<button id="Ok" class="w2ui-popup-btn w2ui-btn">' + options.ok_text + '</button><button id="Cancel" class="w2ui-popup-btn w2ui-btn">' + options.cancel_text + '</button>',
+            onOpen: function () {
+                $('#w2promt').val(options.value);
```
vitmalina: should id be "w2prompt"? (missed p) - in all places

src/w2popup.js
```diff
@@ -989,3 +989,95 @@ var w2confirm = function (msg, title, callBack) {
...
+                $('#w2promt').w2field('text');
```
vitmalina: If it is simple input, there is no need to call w2field('text') on it. However, it might be a good idea to allow user to define this in extended options so that it would be easy to prompt for a number

src/w2popup.js
```diff
@@ -989,3 +989,95 @@ var w2confirm = function (msg, title, callBack) {
...
+                $('#w2ui-popup .w2ui-message .w2ui-btn#Ok').on('click.w2confirm', function (event) {
```
vitmalina: since it is not confirm, should it be "click.w2prompt", I know it does not technically matter, but it will not confuse other people thinking that it somehow related to w2cofirm

vitmalina:
Appreciate your contribution. Frankly, I have never used window.prompt as I never liked how it looks and behaves. Lets see where this will go. Please consider my comments.

src/w2popup.js
antitim: ok

src/w2popup.js
antitim: ok

src/w2popup.js
antitim: yes, my mistake

src/w2popup.js
antitim: ok

src/w2popup.js
antitim: Yes, the idea is interesting. I think to do so: 
```javascript
w2prompt({
  label: 'label',
  field: {
    type: 'type',
    options: {
      prop1: 'prop1',
      prop1: 'prop2'
    }
  }
});
```
What do you think?

antitim:
Thank you, I made a new pull request with fixes.

src/w2popup.js
vitmalina: It would be nice if the options you sent would be same as what w2field takes. Then it would be consistent to configure it.

PR #1177, opened 2016-03-03
Advanced search update

githabas:
1) enum field type in advanced search not worked. I fixed it by commenting part of "switch ($(el).val())".

2) I needed an option to specify custom operators sets in advanced search. So I did it by some invasions in initSearches and getSearchesHTML functions. You can see working examples in test/grid.htm

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -315,17 +315,41 @@
         },
 
         operators: { // for search fields
-            "text"    : ['is', 'begins', 'contains', 'ends'],
-            "number"  : ['is', 'between', 'less:less than', 'more:more than'],
-            "date"    : ['is', 'between', 'less:before', 'more:after'],
-            "list"    : ['is'],
-            "enum"    : ['in', 'not in']
+            "text"    : [['is'], ['begins'], ['contains'], ['ends']],
+            "number"  : [['is'], ['between'], ['less', 'less than'], ['more', 'more than']],
+            "date"    : [['is'], ['between'], ['less', 'before'], ['more', 'after']],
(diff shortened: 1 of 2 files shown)


src/w2grid.js
```diff
@@ -5791,10 +5791,10 @@
             range.hide();
             // fld1.w2field(search.type);
             switch ($(el).val()) {
-                case 'in':
-                case 'not in':
-                    fld1.w2field('clear');
-                    break;
+//                case 'in':
```
mrg2001: Why are these lines commented out, are they no longer needed?

src/w2grid.js
```diff
@@ -315,17 +315,41 @@
         },
 
         operators: { // for search fields
-            "text"    : ['is', 'begins', 'contains', 'ends'],
-            "number"  : ['is', 'between', 'less:less than', 'more:more than'],
-            "date"    : ['is', 'between', 'less:before', 'more:after'],
-            "list"    : ['is'],
-            "enum"    : ['in', 'not in']
+            "text"    : [['is'], ['begins'], ['contains'], ['ends']],
```
mrg2001: I would prefer if you kept the same definition as it was - single array. It is more readable. In case when operator and text for it does not match we can use array. Something like
```
            "text"    : ['is', 'begins', 'contains', 'ends'],
            "number"  : ['is', 'between', ['less', 'less than'], ['more', 'more than']],
```

mrg2001:
Also, in the top of the file, I keep notes what is changed between version. This is then used to update documentation. So, mentioning that search.operators was added will be a good idea.

Issue #1153, opened 2016-02-03
List field ignores readonly/disabled property


mpf82:
When you create a list field and set it to read-only, you can still get the list popup when the field gets the focus (TAB) or when you click in the field and then press arrow up / arrow down key.

![2016-02-03 13_31_07](https://cloud.githubusercontent.com/assets/10152404/12782298/81bd5850-ca7a-11e5-8527-874b24fed682.png)

Code to reproduce the problem:

    <input type="list">
    <input type="list" readonly="readonly">
    <input type="list" disabled="disabled">

    <script type="text/javascript">
    var people = ['George Washington', 'John Adams'];
    $('input[type=list]').w2field('list', { items: people, disabled: true, });
    // $('input[type=list]').prop('readonly', true);
    </script>

It does not matter if readonly/disabled is set on the element or set on run-time.

I managed to track the problem down: the keyUp() handler does not check the properties!

Original code:

        keyUp: function (event) {
            ...
            if (['list', 'combo', 'enum'].indexOf(this.type) != -1) {
                // need to be here for ipad compa
                ...

Suggested fix:

        keyUp: function (event) {
            ...
            if (['list', 'combo', 'enum'].indexOf(this.type) != -1) {
                if ($(obj.el).prop('readonly') || $(obj.el).prop('disabled')) return;
                // need to be here for ipad compa
                ...

**Warning:**
There are other places in the code where you're checking for only readonly **attribute**, when you should probably be checking for both readonly _and_ disabled **properties**:
( For example in keyDown() handler )
I therefore recomment to replace all occurences of 

                if ($(obj.el).attr('readonly')) return;
with

                if ($(obj.el).prop('readonly') || $(obj.el).prop('disabled')) return;

this might be related to issue #244

vitmalina:
Can you please specify what version of the code the problem is. Is it in the current master?

mpf82:
Yes, it's the latest 1.5 version.

mpf82:
While browsng through your site, I found a jsfiddle that you created [here](http://w2ui.com/web/docs/w2form.fields) to demonstrate disabling fields (using version 1.5).

http://jsfiddle.net/rLLwhrbj/1/

You can reproduce the problem in your fiddle by disabling the field and then press arrow up / arrow down while the disabled field has the focus.

mpf82:
added a pull request: https://github.com/vitmalina/w2ui/pull/1188

mrg2001:
Looks like it is fixed

[...]

PR #1188, opened 2016-03-16
 Fixed datetime, fixed readonly, fixed operatorsMap 

mpf82:
Fixed datetime search in grid,
introduced w2utils.settings.datetimeFormat,
accessing readonly / disabled is now done with prop() instead of attr(),
fixed a bug where you could select items from a disabled list - see
issue #1153,
operatorsMap for date, time and datetime now points to "date" instead of
"number"


src/w2utils.js
```diff
@@ -247,7 +248,7 @@ var w2utils = (function ($) {
     }
 
     function isDateTime (val, format, retDate) {
-        if (format == null) format = w2utils.settings.dateFormat + '|' + w2utils.settings.timeFormat;
+        if (format == null) format = w2utils.settings.datetimeFormat;
```
mrg2001: Why not use dateFormat + timeFormat? Less definitions.

mpf82: Users might want to have a different format for date, time and timetime. for example they could want to show seconds on time, but not on datetime.
Or the other way around, they might need a higher precision on datetime, but not on time.

mrg2001: Sure, but it is default format. There is always a way to define custom for any particular case. Anyway, I think it is not a big issue. The pull request is merged. Thank you for your contribution.

mpf82: You're welcome.

Could you please regenerate the files in the dist folder?

PR #1191, opened 2016-03-18
Added column layout to form auto template

mpf82:
Added column layout to form auto template,
added example for columns in form auto template,
updated form fields documentation,
optimized offset access in checkIfMoved() in w2utils


mrg2001:
Seems to be a solid piece of work. Will check the example you created.

PR #1198, opened 2016-03-27
w2tag can be positioned with an array of valid values

mpf82:
w2tag can be positioned with an array of valid values,
added missing German translations


mpf82:
Let me explain this change with a visual example:

2016-03-27 17_14_21-nemetris web template - v0 5

As you can see, a tooltip on a right-aligned tolbar button will go off screen.

However, with this tooltip setting, w2utils will try to position the element so it's completely visible on the screen. If the w2tag can't be displayed completely visible, the last value will be used:

    tooltip: ['top', 'bottom', 'left', 'right', 'top']

2016-03-27 17_14_43-nemetris web template - v0 5

Of course you can also influence w2tag directly:

    $('#my_field_with_w2tag').w2tag("This is a short tag", { position: ['top', 'bottom'] });

This change will have no impact on existing code.

ceedriic:
Does that fix the grid headers?
I just looked at the "test/grid3.html" for example, and it is pretty broken (without your patch):

screen shot 2016-04-02 at 12 21 46

mpf82:
@ceedriic: Yes if you would define the grid tooltip as I did in my example above ( `tooltip: ['top', 'bottom', 'left', 'right', 'top']` ) then the w2tag in your screenshot would render to the bottom of the toolbar button instead.

However, if there's another w2ui layout element above your grid, the calculation would currently still think that the w2tag would fit on the screen (because it does not exactly know the visibility within the container) - so there's still room for improvement in such a scenario.

PR #1202, opened 2016-04-01
w2grid.js: add scrollToColumn(field)

ceedriic:
I've a grid where columns represents the calendar days, from a couple months
in the past to a couple months in the future.

I needed a way to position the grid at "today" column when the grid loads,
therefore I've developed the following small utility function to scroll the view
to a specific column programatically.

Feel free to merge if you find that function useful / correct.


mrg2001:
Seems correct. Thanks for your pull request.

PR #1203, opened 2016-04-02
Improve UI when adding record in the background. Reduce refresh() calls.

ceedriic:
The grid.add() function sometimes calls refresh() 3x internally.
This is slow and messes up the user interface if grid.add() function is used in the background.
Specifically, it becomes impossible tu use the toolbar or searches.

To see the problem, just apply the first patch to test/grid3.html, select the new "adds 1 Record/sec"
button, and try to work with the search dialog.

See individual commit messages for details.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -723,7 +723,7 @@
             return null;
         },
 
-        localSort: function (silent, noReset) {
+        localSort: function (silent, noResetRefresh) {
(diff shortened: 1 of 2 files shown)


mrg2001:
I looked through the commit and it seems correct. I see that you created a new method refreshBody(), in order for it to make it to documentation, need to add a comment on top of the file that method was added. I will do that.

Issue #885, opened 2015-05-05
w2grid.stateSave - save column ordering


prosconi:
http://w2ui.com/web/docs/w2grid.stateSave

It would be nice if ``w2grid.stateSave`` would also preserve column ordering when the user re-orders the columns.

Or maybe I'm doing it wrong?

vitmalina:
I agree.

prosconi:
Actually, I figured out that ``stateSave`` does save the order while serializing the array, but ``stateRestore`` does not apply the order of the columns.

mpf82:
I have a fix prepared, that will restore the column order, still waiting until my other pull request is merged.

mrg2001:
Merged all pull requests

mpf82:
Can be closed, implemented in w2ui 1.5, see  #1206

PR #1206, opened 2016-04-02
various fixes

mpf82:
w2fields: taking care of readonly and disabled if type is enum/file,
w2grid: added missing translation, making reorderRows work even if multiSelect is 'false', saving column frozen state, restoring column ordering,
w2utils: fixed a copy&paste mistake


mpf82:
this will also fix https://github.com/vitmalina/w2ui/issues/885

mpf82:
Could you please recompile the files in the dist folder? Thanks in advance.

mrg2001:
Done

PR #1211, opened 2016-04-11
Fix grid.clear() method.

ceedriic:
Reset searchIds in clear(), simplify.


mrg2001:
I am not sure the change is right. I see clear is used in skip function and with proposed change it will not work properly. Also, there is searchReset() which will reset search fields. And also there is reset() which will reset position of the scroller, etc.

PR #1217, opened 2016-04-21
Update w2grid.js

PrimusFighter:
added possibility to get columndata from script, that delivers json records

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -2414,6 +2414,12 @@
                                     this.records.push(data.records[r]);
                                 }
                             }
+                            // columns
+                            if (data.columns) {
+                                for (var r = 0; r < data.columns.length; r++) {
+                                    this.columns.push(data.columns[r]);
+                                }
+                            }
                             // summary records (if any)
                             if (data.summary) {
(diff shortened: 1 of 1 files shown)


src/w2grid.js
```diff
@@ -2414,6 +2414,12 @@
                                     this.records.push(data.records[r]);
                                 }
                             }
+                            // columns
+                            if (data.columns) {
+                                for (var r = 0; r < data.columns.length; r++) {
+                                    this.columns.push(data.columns[r]);
+                                }
+                            }
```
mrg2001: I am not sure I agree with this. You add the same code to onLoad event and you archived the same thing. Also, what about I click reload button, it will create same columns again.

PrimusFighter:
Ok, the refresh is a point. Solved it with replacing the for loop with
`$.extend( true, this.columns, loadedData.columns );`

onLoad also works with that, but why not use a general nativ solution in gridrendering? in 1.4.x it worked!

mrg2001:
Will think about it, but since there is a good work-arround, it is not critical.

PR #1238, opened 2016-05-07
Fix - w2form.render() calls .refresh() twice / popup overview `onKeydown`

valq7711:
**w2form.render() calls .refresh() twice:**
1.  In `else`-block  of  `if (url && this.recid != 0 && this.recid != null) { this.request(); }...` - it's redundant call, `else`-block was deleted
2. As intended using  setTimeout()

P.S.  to see bug just add `onRefresh` to any form: 
<code>onRefresh: function(e){
            e.onComplete = function(){console.log(e);};
            }</code>

**Fix popup overview**
There is a mistake in options:
It should be  `onKeydown` instead of `onKeyboard`

### DIFF
--- a/docs/overview/popup.html
+++ b/docs/overview/popup.html
@@ -51,6 +51,6 @@ options: {
     onMax           : null,    // event when maximized
     onMin           : null,    // event when minimized
     onToggle        : null,    // event when maximized/minimized
-    onKeyboard      : null     // event when keyboard button is clicked
+    onKeydown      : null     // event when keyboard button is clicked
 }
 </textarea>
--- a/src/w2form.js
+++ b/src/w2form.js
@@ -1254,9 +1254,7 @@
             var url = (typeof this.url != 'object' ? this.url : this.url.get);
             if (url && this.recid != 0 && this.recid != null) {
                 this.request();
-            } else {
-                this.refresh();
-            }
+            } 
             // attach to resize event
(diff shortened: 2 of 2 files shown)


src/w2form.js
```diff
@@ -1254,9 +1254,7 @@
             var url = (typeof this.url != 'object' ? this.url : this.url.get);
             if (url && this.recid != 0 && this.recid != null) {
                 this.request();
-            } else {
-                this.refresh();
-            }
+            }
```
mrg2001: I do not understand why you changed it. If URL is defined, then refresh from URL, otherwise, refresh from memory. This change will cause problems if your form is local

docs/overview/popup.html
```diff
@@ -51,6 +51,6 @@
     onMax           : null,    // event when maximized
     onMin           : null,    // event when minimized
     onToggle        : null,    // event when maximized/minimized
-    onKeyboard      : null     // event when keyboard button is clicked
+    onKeydown      : null     // event when keyboard button is clicked
```
mrg2001: Applied it in another commit

src/w2form.js
valq7711: because at the beginning  of .render( )  `var obj = this;` and at the end (line 1265) `setTimeout(function () { obj.resize(); obj.refresh(); }, 150); // need timer because resize is on timer` - i.e. `.refresh()` is executed anyway.
So, there isn't any problem (no?).

Issue #1210, opened 2016-04-10
TAB not working on button


aetna-softwares:
I am using W2UI 1.4.3 and noticed that I can't leave focus from <button> using TAB key.

I found out that it is because of the w2utils.keyboard

After changed the test
```
if ($.inArray(tag, ['INPUT', 'SELECT', 'TEXTAREA']) !== -1) return;
```
by 
```
if ($.inArray(tag, ['INPUT', 'SELECT', 'TEXTAREA', 'BUTTON']) !== -1) return;
```
It works as expected.

mrg2001:
In 1.5.x there is no more w2utils.keyboard. It has been refactored. Did you try if this works with version in current master?

aetna-softwares:
I didn't try because 1.5 is not yet an official production release so I stick to the stable release.

I may give it a try but I can't find documentation or examples for the 1.5 version. Is there something somewhere ?

A part to make sure that the bug does not exist in future version, is it possible to consider apply the fix in stable version ? Because even if 1.5 is released soon, I think many of us will keep using the 1.4 for a while

mrg2001:
All docs for 1.5.x in this repository in docs sections. Same with demos. As for 1.4.x branch, submit a pull request, I will merge it.

aetna-softwares:
I checked it OK on v1.5

I will read the github doc about how to do pull request ...

aetna-softwares:
Hi @mrg2001 ,

I created the pull request.

mrg2001:
merged

PR #1241, opened 2016-05-11
Issue #1210 fix TAB key handling on BUTTON elements

aetna-softwares:


mrg2001:
I see it was done on 1.4.x branch. There is no plans for new update in that branch.

PR #1249, opened 2016-05-17
fix toolbarColumns bug

SIkebe:
Hidden columns are shown in the toolbarColumns because of "col.hideable".
So I changed it to "col.hidden" and fixed judgement.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -4827,7 +4827,7 @@
             for (var c = 0; c < this.columns.length; c++) {
                 var col = this.columns[c];
                 var tmp = this.columns[c].caption;
-                if (col.hideable === false) continue;
+                if (col.hidden === true) continue;
                 if (!tmp && this.columns[c].tooltip) tmp = this.columns[c].tooltip;
                 if (!tmp) tmp = '- column '+ (parseInt(c) + 1) +' -';
                 col_html +=


mpf82:
This is not a bug and does not need fixing. `hideable` indicates if a column **can** be hidden, not if a column is hidden.

It's even documentd here: http://w2ui.com/web/docs/w2grid.columns

SIkebe:
I see how it is.
So if a column is hidden, it is shown in the toolbarColumns, isn't it?
I think hidden columns should not be shown there...
Could you tell me why hidden columns are shown there by default?

mpf82:
>  So if a column is hidden, it is shown in the toolbarColumns, isn't it?

If `toolbarColumns`is `true`, both hidden **and** visible columns are part of the menu.

>  I think hidden columns should not be shown there...

Why?

>  Could you tell me why hidden columns are shown there by default?

So you can show hidden columns again if you've hidden them.

If you want to hide columns from your user and you don't want the user to be able to make them visible, you could set `hidden` to `true` and `hideable` to `false`

mrg2001:
Not a bug. @mpf82 is right. The toolbar menu purpose is to allow user to hide or show columns. However, if you want a column that is not in the list, then you use hideable = false;

Issue #1251, opened 2016-05-20
Unnecessary resize() and refresh() in form.render() ?!?


mpf82:
In w2form.js (current master) we have this line in the form's render function:

`setTimeout(function () { obj.resize(); obj.refresh(); }, 150); // need timer because resize is on timer` 

I don't think it's really necessary, since we're already doing a resize() 12 lines above the mentioned line, and we're also already doing a request()/refresh() 7 lines above this line, too.

I've removed the line in my local environment and could not see any side effects.

On the contrary, removing this line also removes a possible bug/problem, where the form would try to refresh the fields, even though the form is already destroyed - which occasionally could happen in our application, if the user is browsing through the records really fast (in which case we destroy and re-create the form in fast succession - before the mentioned timer executes - causing a lot of console messages, such as: `ERROR: Cannot associate field "email" with html control. Make sure html control exists with the same name.`).

Before I create a pull request to remove this line, I would like some feedback, as to why it's there in the first place.

If we keep this line, we should somehow prevent the timer from executing if the form is destroyed in the meantime.

valq7711:
You're right - there is something [wrong](https://github.com/vitmalina/w2ui/pull/1238)! If you're sure that `refresh()` is unnecessary  after `request()` I vote for your variant

mrg2001:
submit your pull request, please

mpf82:
PR created: https://github.com/vitmalina/w2ui/pull/1254

PR #1254, opened 2016-05-23
Minor improvements

mpf82:
added support for 'accept' attribute for file type, 
removed setTimeout() call in w2form ( see https://github.com/vitmalina/w2ui/issues/1251 )


src/w2form.js
```diff
@@ -1264,7 +1264,6 @@
                 this.tmp_resize = function (event) { w2ui[obj.name].resize(); };
                 $(window).off('resize', 'body').on('resize', 'body', this.tmp_resize);
             }
-            setTimeout(function () { obj.resize(); obj.refresh(); }, 150); // need timer because resize is on timer
```
mrg2001: Are you sure this needs to be removed? I fail to see how it relates to accept attribute

mrg2001: I will run my test and see if I can find anything.

mpf82:
Removing this line is unrelated to the accept attribute.

It's related to #1251 where I stated that this line is unneccessary and you said I should create a pull request.

PR #1261, opened 2016-05-27
Fixed #1258 - add fewer empty-records

### LINKED ISSUE #1258: w2grid: possible to scroll empty records, hiding actual content
w2ui 1.5

https://jsfiddle.net/062rytft/1/

On browsers/operating systems that allow it, it is possible to vertically scroll a grid that has no vertical scrollbar by using the middle mouse button. This seems to be due to the `.w2ui-empty-record`s that w2ui adds to grids without enough content to fill themselves in order to keep the styling consistent. W2ui adds a few more than just the height of the grid and since the browser sees them as content it is possible to scroll through them in certain scenarios. This issue only occurs when horizontal scrollbars are present.

Additionally, this presents an issue where if you have a grid with a lot of records, scroll to the bottom of it, replace the records array with a smaller data set, and then refresh the grid, the top few, or potentially all, records will be cut off with no indication that they exist.

This can be seen in the above fiddle if you:

1. Click the 'Add records' button
2. Scroll to the bottom of the table
3. Click the 'Change records' button

Notice how the grid appears completely empty, with no vertical scrollbar visible and no way to scroll up with the scrollwheel, despite having a record in it. This only happens when manually changing the `.records` array and refreshing - `.add` `.remove` and `.clear` work fine. Again, this issue only occurs when horizontal scrollbars are present.

Both issues were reproducible on Windows 7 with Chrome 50.0.x, only the second was reproducible on Firefox 46.0.1.

swarner0807:
My attempt at a fix for #1258. Changed so that only the exact number of empty records required to fill the height of the grid are created, rounded down. Any leftover space is filled with another empty record of partial height. This way there is no extra "content" that can be accidentally scrolled in certain situations.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -5549,30 +5549,40 @@
             }
             frecords.css({ overflow: 'hidden', top: records.css('top') });
             if (this.show.emptyRecords && !bodyOverflowY) {
-                var max = Math.floor(records.height() / this.recordHeight) + 1;
-                if (this.fixedBody) {
-                    for (var di = buffered; di <= max; di++) {
-                        var html1 = '';
-                        var html2 = '';
-                        var htmlp = '';
-                        html1 += '<tr class="'+ (di % 2 ? 'w2ui-even' : 'w2ui-odd') + ' w2ui-empty-record" style="height: '+ this.recordHeight +'px">';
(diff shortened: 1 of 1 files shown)


swarner0807:
Sorry about that, not sure how that got past me! I just put out another PR that should fix this issue. Thank you for reporting it!

PR #1263, opened 2016-05-30
Fixed error caused by PR #1261

swarner0807:
Fixes a small error in PR #1261. Now makes a check to ensure `records[0]` is not undefined before getting its `scrollHeight`.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -5550,7 +5550,7 @@
             frecords.css({ overflow: 'hidden', top: records.css('top') });
             if (this.show.emptyRecords && !bodyOverflowY) {
                 var max      = Math.floor(records.height() / this.recordHeight) - 1;
-                var leftover = records[0].scrollHeight - max * this.recordHeight;
+                if (records[0]) var leftover = records[0].scrollHeight - max * this.recordHeight;
                 if (leftover >= this.recordHeight) {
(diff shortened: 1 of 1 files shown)


mpf82:
frecords.css({ overflow: 'hidden', top: records.css('top') });
             if (this.show.emptyRecords && !bodyOverflowY) {
                 var max      = Math.floor(records.height() / this.recordHeight) - 1;
    -                var leftover = records[0].scrollHeight - max * this.recordHeight;
    +                if (records[0]) var leftover = records[0].scrollHeight - max * this.recordHeight;
                 if (leftover >= this.recordHeight) {
                     leftover -= this.recordHeight;
                     max++;

Woudn't this result in `leftover` being **undefined** if records has no items, which means the next lines would crash in that case?

ceedriic:
This is better, the test does not crash anymore.
But it calls addEmptyRow with an undefined height, which adds the following html junk:

`<tr class=\"w2ui-even w2ui-empty-record\" style=\"height: undefinedpx\"><td class=\"w2ui-col-number\"></td><td class=\"w2ui-grid-data-last\"></td> </tr>`

swarner0807:
Good point. It technically works the way this is used, but it would be much safer and cleaner to make sure `leftover` is defined regardless. I have updated the PR. Thanks!

PR #1264, opened 2016-05-30
Grid: make 'editable' selection more flexible 

ceedriic:
Make it easy to define a cell editor which is different for each cell, not only per-column.

This is done by allowing to optionally define col.editable as a function. In this situation, when w2ui need to check if and how a given cell can be edited, it will call col.editable() with the same parameter than col.render(), and col.editable() shall return null (cell not editable) or an appropriate 'editable' object.

This patch add a 'getCellEditable()' method, with similar argument than getCellValue(). getCellEditable() will look at col.editable and rec.w2ui.editable and return null (cell not editable) or the right 'editable' object descriptor. The grid code is changed to use getCellEditable() consistently everywhere when the "editable" state of a cell is checked.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -116,6 +116,8 @@
 *   - message(.., callBack) - added callBack
 *   - grid.msgEmpty
 *   - field.render(..., data) -- added last argument which is what grid thinks should be there
+*   - col.editable can be a function which will be called with the same args as col.render()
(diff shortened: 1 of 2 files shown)


mpf82:
`if (edit == false) return null;` 

this should probably be:

`if (edit === false) return null;` 

because **undefined** or **null** will implicitly evaluate to **false** if you use `==` instead of `===` and the function would return.

ceedriic:
Oops, yes, you're right. I will fix that shortly. Thanks.

PR #1269, opened 2016-06-19
w2form and typos

mpf82:
w2form's tabs can now be active by default by adding "active: true" property,
fixed refresh when value is empty or zero,
reduced flickering when form has multiple pages


src/w2form.js
```diff
@@ -995,7 +1002,7 @@
                 if (tmp) tmp.clear();
                 $(field.$el).off('change').on('change', function () {
                     var value_new      = this.value;
-                    var value_previous = obj.record[this.name] ? obj.record[this.name] : '';
+                    var value_previous = obj.record[this.name] !== undefined ? obj.record[this.name] : '';
```
mrg2001: I prefer in this case to use "!= null" (single equal). Which is true when variable is null or undefined.

PR #1281, opened 2016-06-27
Create en-id.json

hassanarrizal:
Indonesia localization


mpf82:
You have translated "All Fields" twice. 

Once as `"All Fields": "Tambah Kolom",` and at the end as `"All Fields":"Semua Kolom"`.

mpf82:
There's no string `"Add Fields"` in all of w2ui, you chould have just deleted the second `"All Fields"` ;)

PR #1291, opened 2016-07-08
Change to avoid columns to shrink to the left on grid selection.

rcpacheco:
The current implementationis causing the columns to shrink to the left, leaving an empty section to the right.
So, the grid must gain focus before any change in its left  and top values. I left it inside a setTimeout of 1, because previous version used it also.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -4603,32 +4603,33 @@
                     if (obj.last.move.recid == null) obj.last.move.type = 'select-column';
                     // set focus to grid
                     var target = event.target;
-                    var $input = $(obj.box).find('#grid_'+ obj.name + '_focus');
-                    // move input next to cursor so screen does not jump
-                    if (obj.last.move) {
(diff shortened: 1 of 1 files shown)


mrg2001:
This code breaks previous functionality. Here is how to repeat it.

1. Create a grid that is greater in height then browser windows. And add to it many records so that scrollbars in the grid and browser are shown.
2. Scroll down both browser scroll bar and grid scrollbar
3. Click on the record in the grid.

What happens, browser scrollbar jumps up. It remember the reason it jumps because the focus input gets focus and browser moves scroll bar to see that element.

Please give me steps how to repeat your bug and I will look at it.

mrg2001:
I think I found out how to repeat and fix shift left problem, here is the fix https://github.com/vitmalina/w2ui/commit/9c8d2408d0a8eb5f59277b7227a5a4c464e59fd9

rcpacheco:
I applied the latest update and it works. Thanks a lot.

PR #1346, opened 2016-10-04
translating input list value after select an item with translation

daniel-alves:
1. As few as was added to translate the text into the input when the item is selected, because the drop-down menu was translated and after selecting the input not translated.


mpf82:
Files in the `/dist` folder are generated automatically from the files in the `/src` folder. 

You're not supposed to edit/submit them, as the changes would be lost anyway when the files are rebuilt.

dist/w2ui-fields.js
```diff
@@ -3475,7 +3475,7 @@ w2utils.event = {
                     var focus = obj.helpers.focus.find('input');
                     if ($(focus).val() === '') {
                         $(focus).css('text-indent', '-9999em').prev().css('opacity', 0);
-                        $(obj.el).val(selected && selected.text != null ? selected.text : '');
+                        $(obj.el).val(selected && selected.text != null ? w2utils.lang(selected.text) : '');
                     } else {
```
mrg2001: selected.text comes from the list of options and it should already be translated.

dist/w2ui.js
```diff
@@ -18930,7 +18933,7 @@ var w2prompt = function (label, title, callBack) {
             }
             // focus on load
             function focusEl() {
-                var inputs = $(obj.box).find('input, select, textarea');
+                var inputs = $(obj.box).find('div:not(.w2ui-field-helper) > input, select, textarea').not('.file-input');
```
mrg2001: will add

PR #1358, opened 2016-10-27
[w2grid]: trigger before/after events for all commands - Added calendar support to DatePicker and also Jalali calendar

afz:


mrg2001:
Thanks for you contribution. However, I have some questions.

1. Please note that there is a settings in w2utils.settings that defines if week starts with Monday or Sunday. This is deemed to cover all use-cases in the world. There is no need to do any updates in locale files. 
2. I think it would be best to follow tightly with what JavaScript date Object does. And the attempt has been make to use it to calculate dates. For example, to add 5 days to a specific date, I would convert it to unix time (with Date.getTime() function) add number of milliseconds that constitute 5 days (5 * 24 * 60 * 60 * 1000) and then construct JavaScript date object based on it and convert it to date.
3. Also, please see unit tests for date conversion in /qa folder

Please describe your motivation for this changes.

mrg2001:
Since there is no response for my questions. I am closing this pull request.

PR #1376, opened 2016-11-08
Update w2grid.js - Search support for date operators between, less and more for CE formats

ecivic:
Search support for date operators between, less and more for CE formats like Germany, Croatia, etc. Tested with default and custom locale (dd.mm.yyyy).


mrg2001:
Thanks for you contribution.

PR #1394, opened 2016-11-24
Various fixes

mpf82:
added jquery dependency to package.json,
updated readme.md,
added support for form fields without label (.w2ui-span0),
fixed CSS position of elements in toolbar search input when scrolling toolbar (left/right),
fixed inconsistent usage of apostrophes,
fixed sort direction support for NULL value and mixedcase/uppercase values


mpf82:
missing a comma in ``package.json``, but it seems I'm not able to create another PR while this one is still pending.

mrg2001:
I will fix the coma issue

PR #1396, opened 2016-11-28
Charts with d3.js

MaximPetrichuk:
Hi!
Propose add a new object to the library.

Regards, Maxim.

### DIFF
--- a/demos/examples/chart-1.html
+++ b/demos/examples/chart-1.html
@@ -1,0 +1,69 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>W2UI Demo: chart-1</title>
+    <script src="http://ajax.googleapis.com/ajax/libs/jquery/2.1.0/jquery.min.js"></script>
+    <script src="https://d3js.org/d3.v4.min.js"></script>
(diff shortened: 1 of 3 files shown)


mpf82:
I like the idea.

Btw, in the "prototype", you have declared ``this.last.g`` twice. The second one (line 41) should probably be ``g1``.

MaximPetrichuk:
Sorry, right - should be g1.

MaximPetrichuk:
I'm new in Githab. If changes after PR but before its adoption, it is necessary to cancel and create a new PR?

mrg2001:
I am not 100% sure about charts, as I feel it deserves its own repository. I have worked with d3 extensively and do have some samples too. Let's give it a try and see where it will lead. Should it stay as a single w2ui.js file or be split into its own file, not sure yet.

MaximPetrichuk:
Thank, this is my first pull request :)

aadrian:
> I am not 100% sure about charts, as I feel it deserves its own repository.

+1 for their own repository (at least what NPM and Bower is concerned), i.e. W2UI should include only the required changes (if any) to allow charts easily, but no dependencies or charts IMHO.
*(many already use the one or the other charting library, so this extension would be useful more for **new** projects that haven't decided upon a charting lib yet)*

The Demos could include the charts too, since it makes no sense to host many different demos.

PR #1437, opened 2016-12-29
CommonJS/AMD support

giladno:


mrg2001:
Files in /dist/* should not be submitted. I looked through the change and it looks small. I will test it and if it does not break anything for people who do not use commonJS, I will merge it.

mrg2001:
Please resubmit without dits/* files

PR #1464, opened 2017-01-12
Update w2grid.js

araujomario:
The "textSearch" default value shall be "contains" and not "begins" in order to maintain compatibility with previous versions.


mrg2001:
Well, I feel it is better to have "begins" as default search operator for text. However, when you define search field, you can pass operator, so this should not cause problems.
```js
searches: [
 { type: 'text', caption: 'Text', field: 'fname', operator: 'contains' }
...
]
``

Issue #1466, opened 2017-01-13
w2field list doesn't distinguish item ids when id = '' and id = 0


avber:
comboItems.push({ id: '', text: 'Select' })

comboItems.push({ id: 0, text: 'Select 2' })

they are treated incorrectly after selection

[http://jsfiddle.net/rstv96s0/](http://jsfiddle.net/rstv96s0/)

mendozada:
Hello, the structure of the item object, requires that the ID be a unique key,

[http://w2ui.com/web/docs/1.5/form/fields-list](url)

avber:
IDs ARE unique: '' and 0

mpf82:
I created a PR to fix this issue: https://github.com/vitmalina/w2ui/pull/1477

Reason why it's currently not working: The ``==`` operator in JS compares for ``truthy`` and ``falsy`` instead of exact matches. Fix: use ``===`` operator instead, which is also type-safe.

avber:
@mpf82 
Yeah, I know about  === operator :)

If you have time, could you also create a PR for #1421 ?

Thanks

mpf82:
To be frank, I have no idea how to create a second PR in the github desktop app, while I still have an active/pending PR.

mpf82:
PR has been merged, please close this issue, unless you can still reproduce the problem with the latest sources.

PR #1477, opened 2017-01-19
Fixed issue #1466

mpf82:
Fixed issue #1466


gamedevsam:
This seems like a safe, straightforward change.

Issue #1493, opened 2017-02-03
Minor regex fix for isEmail check


stmansour:
Hi - the check did not allow email names with dashes. This is a valid character for the name (started receiving a bunch of them, that's when we found the issue). 

I changed this:
    var email = /^[a-zA-Z0-9._%-+]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/;
to this
    var email = /^[-a-zA-Z0-9._%-+]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/;

Now it works for email addresses like ys-k37@outlook.com.  This is a very quick fix.  But I can do a pull request if that helps.

-S

Supme:
and domain, for example "@dmbasis.agency", have more then 4 litters
and domain "@тюменев.рф"...

Supme:
Best way convert domain to punycode and after validate this

Issue #1404, opened 2016-12-01
Grid reloads from first record on last "get" even though record count doesn't change


stmansour:
By default the grid asks for batches of 100 records at a time. When I scroll past the first 100, it sends a request for the next 100, and so on.  I've tested this capability and it works fine until it gets to the last batch. When the last batch is delivered, it pops up a dialog that says "Your remote data source record count has changed, reloading from the first record."  But the record count did not change.  I put some logging into the server to validate. Here is the output for a test where there were 427 records.  You can see the five requests the grid sent to the server. I list how many records (transactants) were read from the database for each query (stored in the records array), and also the first 50 chars of the response sent back to the grid.  Note that when the last 27 records are sent, total is 427, same as with all the previous replies.  I checked the docs but I didn't see anything about handling the last batch different.  So, I'm assuming there's some sort of logic error in the grid code on the last batch.

grid request:  request={"cmd":"get","selected":[],"limit":100,"offset":0}
Loaded 100 transactants
first 50 chars of response: {"status":"success","total":427,"records":[{"recid

grid request:  request={"cmd":"get","selected":[],"limit":100,"offset":100}
Loaded 100 transactants
first 50 chars of response: {"status":"success","total":427,"records":[{"recid

grid request:  request={"cmd":"get","selected":[],"limit":100,"offset":200}
Loaded 100 transactants
first 50 chars of response: {"status":"success","total":427,"records":[{"recid

grid request:  request={"cmd":"get","selected":[],"limit":100,"offset":300}
Loaded 100 transactants
first 50 chars of response: {"status":"success","total":427,"records":[{"recid

grid request:  request={"cmd":"get","selected":[],"limit":100,"offset":400}
Loaded 27 transactants
first 50 chars of response: {"status":"success","total":427,"records":[{"recid

mrg2001:
no longer an issue, see https://github.com/vitmalina/w2ui/pull/1521

Issue #1401, opened 2016-12-01
Grid does not empty when search results from server indicate no matches


stmansour:
Hi,
Grid is working fine with server searches except when there are no matches.  In the w2ui demos where the data records are embedded in the JS, if the search string produces no matches then the grid empties. That is what I would expect. However, when I do a search against a server and there are 0 results, the grid does not empty.  It retains the contents that it had prior to the search.  Here is the json the server returns when there are no matches:

  {"status":"success","total":0,"records":null}

The JS editor in Chrome indicates the problem 
![image](https://cloud.githubusercontent.com/assets/6549488/20782290/6914e496-b73f-11e6-8801-dcc9977221a6.png)
-- an attempt to read the length on a null value.  I'm hoping you can just check for records being null and do the right thing -- the json above was generated by Go's json encoder; meaning that if you require something different in the json then Go programmers would need to special-case the response for 0 matches.

I made a fix that seems to work.  In w2ui-1.5.rc1.js  I changed this:

5321                             if (data.records.length == this.limit) {
5322                                 this.last.xhr_hasMore = true;
5323                             } else {
5324                                 this.last.xhr_hasMore = false;
5325                                 this.total = this.last.xhr_offset + data.records.length;
5326                             }

to this:

5321                             if (data.records != null && data.records.length == this.limit) {
5322                                 this.last.xhr_hasMore = true;
5323                             } else {
5324                                 this.last.xhr_hasMore = false;
5325                                 this.total = this.last.xhr_offset; 
5326                                 if (data.records != null) this.total = data.records.length;
5327                             }

-Steve

mpf82:
To me it sounds like a bad server-side implementation. Instead of returning ``null`` the server should return an empty array.

Still, nothing wrong with w2grid doing some checks instead of assuming that properties are not null. Maybe even check with ``if(Array.isArray(data.records))``, to make sure records aren't of any other type, even if not null.

stmansour:
I agree that "records": null could be replaced with an empty list -- that would probably keep a lot of code from breaking.  This is just the behavior of Go's json encoder.  I'll mention this to the go community.

PR #1521, opened 2017-02-16
Fixes for #1493,#1404,#1401

stmansour:
1493 - fixes a minor regexp issue that caused email addresses with a dash (-) in the username to fail validation.
1404 - is no longer an issue with the fix I've included for 1401 - it was a problem I introduced with my initial fix for 1401
1401 - Fixes the issue caused by golang's default json marshaler. Sometimes the record set for a search is empty and golang's json marshaler would omit 'records' completely from the output stream rather than supplying something like  "records": []


dist/w2ui-fields.js
```diff
@@ -156,7 +156,7 @@ var w2utils = (function ($) {
     }
 
     function isEmail (val) {
-        var email = /^[a-zA-Z0-9._%-+]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/;
+        var email = /^[-a-zA-Z0-9._%-+]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/;
```
mrg2001: no need to change this file as it is generated

PR #1528, opened 2017-02-21
Adding "hidecheckbox" property to a w2grid record

alexey-vasyunin:
This property will hide checkbox in record in grid with selectColumn. If w2ui.GRID.records[N].w2ui.hidecheckbox set true the checkbox in this row will be hidden. It's helpful when the row has subgrid or childern records (in a tree-like grid) and need to allow to select only start branches of tree.


mpf82:
Please note that ``w2ui`` is an **optional** record property. Your code will throw an error if the record does not contain this property!

alexey-vasyunin:
Thanks @mpf82 , you right. I've fixed and tested it.

mrg2001:
I am going to merge, but will change to camelCase

PR #1544, opened 2017-03-07
Added docs for textSearch and httpHeaders

mpf82:
Added docs for textSearch and httpHeaders


mpf82:
@mrg2001 hi! you merged this change and I can see it in the repository - however it seems it does not show in the online documentation ( http://w2ui.com/web/docs/1.5/grid/properties ) - are there other files that I must modify, so the updated docs show up in the online docs?

PR #1557, opened 2017-03-30
bugfixes,  improvements

lacmuch:
sry i pulled dists :(


src/w2utils.js
```diff
@@ -1822,7 +1822,7 @@ w2utils.formatters = {
 
     'money': function (value, params) {
         if (value == null || value === '') return '';
-        var data = w2utils.formatNumber(Number(value), w2utils.settings.currencyPrecision || 2);
+        var data = w2utils.formatNumber(Number(value), w2utils.settings.currencyPrecision===0?0:w2utils.settings.currencyPrecision || 2 );
```
mrg2001: spaces needed before and after ====, ., :

ms264556:
There are some really useful changes here. Do you know why they are not being committed? For example, I have a form resizing bug which is 'fixed' by applying and using c4de012.

vitmalina:
Thank you for your contributions. I think our branches diverged too much. I think it will not be possible to merge them. Thanks.

PR #1590, opened 2017-05-03
added support for cyrillic emails

antitim:
For validation of email addresses such as test@домен.москва


mrg2001:
No need to push files from /dist folder, they are all generated

PR #1605, opened 2017-05-24
Fixed some documentation paths

edomato:
I fixed the path of the "href" so they work when you browse the documentation using an webserver (for example, the internal webserver of PHP).

Maybe we could add some instructions somewhere to explain how to use the documentation.

Greets


mrg2001:
I found why it was this way. When I deploy to w2ui.com it needs ../, but for local it does not. There is a script that generates summaries and it is located in docs/generate.js. It will generate either local or for w2ui.com

```js
cd docs
node generate.js
```

Will regenerate all summary pages based on documentation. So, there is not need to change these files manually.

PR #1607, opened 2017-05-27
w2grid: optimise localSort()/localSearch() and the get() function

ceedriic:
A customer reported that a local sort operation involving ~35K records in a tree took many *minutes* to complete. It turns out that the root cause is the get() function being slow, and because get() is called for each element in localSort() => preparePaths() => getRecordPath(), this lead to quadratic behaviour O(N^2).

This patch optimize w2grid.get() and turn it from O(N) to O(1) by introducing a cache. The search operation that took minutes now complete in a couple of seconds.

Note: the next functions that need to be optimised to make that operation faster in my test are isDate() and prepareRecord(). I will try to take a look at them next month if possible.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -507,34 +507,42 @@
             // search records
             if ($.isArray(recid)) {
                 var recs = [];
-                for (var i = 0; i < this.records.length; i++) {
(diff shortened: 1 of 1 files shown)


mrg2001:

PR #1608, opened 2017-05-27
w2grid: make splitting of field name optional

ceedriic:
Introduce useFieldDot configuration option to make the field splitting
behaviour optional.

This is useful to turn this off if field names could contains dots.


mrg2001:
As long as by default it will use . as object notation, I do not might if we have option not to do so.

PR #1642, opened 2017-07-26
bugfix find() method - return array of record objects

agata-maksymiuk:
Bug in w2grid.find metod:
Returns array of ints when returnIndex set to true, instead array of record objects.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -474,7 +474,7 @@
                     }
                 }
                 if (match && returnIndex !== true) recs.push(this.records[i].recid);
-                if (match && returnIndex === true) recs.push(i);
+                if (match && returnIndex === true) recs.push(this.records[i]);
             }
             return recs;
         },


src/w2grid.js
```diff
@@ -474,7 +474,7 @@
                     }
                 }
                 if (match && returnIndex !== true) recs.push(this.records[i].recid);
-                if (match && returnIndex === true) recs.push(i);
+                if (match && returnIndex === true) recs.push(this.records[i]);
```
vitmalina: Not sure it is right. If returnIntex is true, then it should return index of the record that is found.

agata-maksymiuk: Then documentation is wrong:
http://w2ui.com/web/docs/1.5/w2grid.find
It says "boolean, if true returns records not their recids"
And also the way it works as described in documentation is much more useful (to get whole record object than just array index which means and does nothing).

vitmalina:
Fixed documentation, will upload later

PR #1692, opened 2017-10-17
Dragging a column to reorder can result in a ghost column you can no longer interact with.

codertd:
Reproducible by left clicking a column and starting to drag it, then quickly releasing click and re-clicking. Selecting text first may or may not help. (seems to?)  This fix looks for an attribute on the click event that isn't present under normal circumstances. 

If the attribute is found the handler just returns, and W2ui wont allow you to release the column, so it just continues to be under your mouse until you place it where it needs to go to be reordered, etc.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -5203,7 +5203,15 @@
                 _dragData.pressed = true;
 
                 _dragData.timeout = setTimeout(function(){
-                    if ( !_dragData.pressed ) return;
+                    // When dragging a column for reordering, a quick release and a secondary
+                    // click may result in a bug where the column is ghosted to the screen,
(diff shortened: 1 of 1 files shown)


src/w2grid.js
```diff
@@ -5203,7 +5203,15 @@
                 _dragData.pressed = true;
 
                 _dragData.timeout = setTimeout(function(){
-                    if ( !_dragData.pressed ) return;
+                    // When dragging a column for reordering, a quick release and a secondary
```
vitmalina: I think it is too excessive use of comments. If you would like to add so much comments, create an issue on github and put a link to the issue in comments. I will remove them later on.

Issue #226, opened 2013-08-29
suggestion for locking


rolfn:
w2form has a lock function for the whole form. Is it possible to add separate lock functions for each tab?

...Rolf

edomato:
This is something interesting to have, and I'm also in the need of other kind of locking.

What I'm needing (that right now I'm implementing outside W2UI but hope to make a pull request soon) is the possibility to disable all the form fields so in that way, if you have a form with tabs for example you can still visit each tab, view the content of the whole form and not being able to modify it.

Maybe this solution could be extended to the tab locking feature requested here, so you can disable (or enable) just some tabs of a form.

Just my two cents 😃

mpf82:
implememtend in PR #1704 - this issue can be closed

PR #1704, opened 2017-11-02
w2form lockPage and unlockPage

mpf82:
w2form lockPage and unlockPage, see issue #226


src/w2form.js
```diff
@@ -789,6 +789,28 @@
             setTimeout(function () { w2utils.unlock(box, speed); }, 25); // needed timer so if server fast, it will not flash
         },
 
+        lockPage: function(page, msg){
```
vitmalina: I usually add in comments on top of the file what was added. It is needed to update documentation later.

Issue #1686, opened 2017-10-10
tabindex problem when two forms in same page


nproto:
* **I'm submitting a ...**
[] bug report
[x] feature request
[ ] question about the decisions made in the repository

* **What is the current behavior?**
When there are two forms autogenerated in the same page the focus from the first field of the first form jumps to first field of the second form, then it comes back to the second of the first form and then again to the second of the second form and so on.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots**.

* **What is the expected behavior?**
The focus  should stay in the same form until all field is exhausted.

* **What is the motivation / use case for changing the behavior?**
There many cases where two forms are present in the same page

* **Link to jsFiddle that demonstrates the problem:**

* **Other information**
The problem occurs because the tabindex numbering of the fields  starts from 1 in each form. So there are two fields with tabindex 1, two with tabindex 2 and so on.
A solution could be to allow the setting of tabindex in the field options or even set the starting number of the autoincrement used for the tabindex, in the form options.

nproto:
I noticed that I can set the tabindex through the attr  option. When I set for example 
{ field: 'ff', type: 'test', html: { caption: 'ff', attr:'tabindex=102' } } it works fine.

mpf82:
I created a PR #1708, that will not completely solve this issue, but you can now assign a value to ``w2form.tabindexBase``, so you do not need to assign a tabindex to every field manually.

nproto:
This will solve the issue. Thank you.

vitmalina:
PR merged

Issue #1660, opened 2017-08-26
What with Raspberries ?


okinadok:
what is the fate of vitmalina ?
what's wrong with him ?
my email : okinadok@gmail.com

mpf82:
@mrg2001 / @vitmalina Not a w2ui issue, this topic can probably be closed.

PR #1708, opened 2017-11-04
added w2form.tabindexBase

mpf82:
added w2form.tabindexBase, added docs, related issue: #1686, ran code through jshint, replaced some == and != comparisons with === and !==

Edit: the first radio item in a radio group will also get a tabindex now.


mpf82:
@vitmalina Did I modify/create the ``/docs`` files correctly? The property and the page content didn't show up for me, but I think that's because the property was not yet in the w2form prototype?

Oh, and maybe you can close the issues that I have linked in my recent PRs?

I have tried to review almost all open issues, trying to fix them or commenting that they can be closed, mostly because of missing clarification/inactivity - or because they do not belong to the issues, such as this topic: #1660

Cheers =)

rsudip90:
Is it released yet (in nightly build)?

rsudip90:
and what about form template, aren't we supposed to set the tabIndex with value for form templates too? Right now it is only being set in `generateHTML` method, instead of that it should be handle inside render or maybe refresh event. So that it would work with auto-generated html code as well as form template's html code? Otherwise `tabindex` with value has to be put manually in form template! :) I don't know but I'm just throwing an idea so that it could be userful and helpful for both option!

PR #1711, opened 2017-11-06
w2uiのアップデート　

takahashi-tosh:
sorry never mind this pull request
pls close


vitmalina:
You should not change files in /dist/* folder as they are generated. Please change the source and then it is compiled with grunt.

PR #1729, opened 2017-12-11
Grid statesave

mpf82:
make grid.stateSave more flexibel


mpf82:
Use case: also save ``column.render`` if you want to.

Default settings will behave the same as the current stateSave().

This makes it also possible to save any custom column property.

PR #1746, opened 2018-02-28
fix resize priority in w2overlay when the search box is opened

vipersss:
when the "searches" contains elements of the type "enum" and "resize()" occurs before the "initSearches()", the "height" break down on div in w2ui-overlay-"name"-searchOverlay

image

### DIFF
--- a/src/w2utils.js
+++ b/src/w2utils.js
@@ -2475,9 +2475,9 @@ w2utils.event = {
 
         // need time to display
         setTimeout(function () {
-            resize();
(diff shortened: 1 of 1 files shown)


vipersss:
https://github.com/vitmalina/w2ui/pull/1746/commits/76e59a023bc14d611b3a3f87ec3cec9031455bb7

When the content in search box changes, height of the container is also changing.
The parent div must change the height
bug:
image

fixed:
image

PR #1754, opened 2018-03-07
Make w2popup also work in a module environment

CherryDT:
Ensure that the `w2popup` object is also exported. Before, only `w2ui`, `w2obj` and `w2utils` were exported.

I moved the module export part to its own file which is included last, instead of adding it to `w2utils`, because a) I think it doesn't really belongs to the utils file and b) It wouldn't be able to export `w2popup` which is declared in another file later on (it would export it as `undefined`).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -76,7 +76,8 @@ module.exports = function (grunt) {
                     'src/w2toolbar.js',
                     'src/w2sidebar.js',
                     'src/w2fields.js',
-                    'src/w2form.js'
+                    'src/w2form.js',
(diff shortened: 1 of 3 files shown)


vitmalina:
Can you please put this code into w2utils. I do like an extra file, as it feels unnecessary.

CherryDT:
No because that is exactly the issue.

The module export needs to be the last file in the list (that was the problem I was fixing in this PR - since it wasn't the last, it was missing some things), but putting w2utils as last file breaks things, and stuffing it in w2form doesn't seem like the right choice...

PR #1776, opened 2018-05-29
Fix "grid reloads after skip and scroll to end"

### LINKED ISSUE #1775: Grid reloads after skip and scroll to end
* **I'm submitting a ...**
[x] bug report
[ ] feature request

* **What is the current behavior?**
[w2grid.skip](http://w2ui.com/web/docs/1.5/w2grid.skip) causes the grid to reload when reaching the end of the list.

* **Steps to reproduce**.
1) Open [infinite scroll demo](http://w2ui.com/web/demos/#!combo/combo-9).
2) Through the second toolbar button (show/hide columns) enter "skip" 3580, so that last request receives less rows than requested.
3) Scroll to end and the grid will reload to initial offset.

image

* **What is the expected behavior?**
It should be possible to reach the end of the list.

* **What is the motivation / use case for changing the behavior?**
A fairly significant problem that should be easy to fix.

* **Other information**
I think [line 2575 in _requestComplete_](https://github.com/vitmalina/w2ui/blob/1de0c6dc7a756561fdb734ae14ba6972805f7b8c/src/w2grid.js#L2575) should be
`this.total = this.offset + this.last.xhr_offset + data.records.length;`
so as to correspond to [line 2388 in _request_](https://github.com/vitmalina/w2ui/blob/1de0c6dc7a756561fdb734ae14ba6972805f7b8c/src/w2grid.js#L2388).

unix786:
Fix calculation of total in requestComplete when there is an initial offset made with skip(). This fixes #1775.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -2572,7 +2572,7 @@
                                 this.last.xhr_hasMore = true;
                             } else {
                                 this.last.xhr_hasMore = false;
-                                this.total = this.last.xhr_offset + data.records.length;
+                                this.total = this.offset + this.last.xhr_offset + data.records.length;
                             }
                             if (this.last.xhr_offset === 0) {
                                 this.records = [];


src/w2grid.js
```diff
@@ -2572,7 +2572,7 @@
                                 this.last.xhr_hasMore = true;
                             } else {
                                 this.last.xhr_hasMore = false;
-                                this.total = this.last.xhr_offset + data.records.length;
+                                this.total = this.offset + this.last.xhr_offset + data.records.length;
```
mrg2001: Thank you for submitting the pull request. However, I am not sure that it is correct. I looked up in the code and I see that this.last.xhr_offset already has this.offset in it. Adding it second time would be a mistake. Also, this condition is only applicable in infinite scroll and when the number of records is not set from the server ahead of time.

Please provide a use case when that shows that this change is good. If you can create a sample service and an example on jsFiddle (or similar) it would be great.

unix786: The number of records isn't always known ahead of time. One of the reasons is that data could still be in the process of loading/generating server-side as the user is scrolling.

An example of this behavior is in "steps to reproduce" in issue #1775 .

unix786:
Moved the commit onto a new branch (off of the main branch).

Messed up the fork at first... not sure if I did it property in the end. Still learning to use Git. So far it's as easy and intuitive as piloting a military helicopter.

PR #1839, opened 2019-06-20
Fix "grid reloads after skip and scroll to end"

### LINKED ISSUE #1775: Grid reloads after skip and scroll to end

unix786:
Fix calculation of total in requestComplete when there is an initial offset made with skip(). This fixes #1775 .

Supersedes PR #1776 .

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -2572,7 +2572,7 @@
                                 this.last.xhr_hasMore = true;
                             } else {
                                 this.last.xhr_hasMore = false;
-                                this.total = this.last.xhr_offset + data.records.length;
+                                this.total = this.offset + this.last.xhr_offset + data.records.length;
                             }
                             if (this.last.xhr_offset === 0) {
                                 this.records = [];


src/w2grid.js
```diff
@@ -2572,7 +2572,7 @@
                                 this.last.xhr_hasMore = true;
                             } else {
                                 this.last.xhr_hasMore = false;
-                                this.total = this.last.xhr_offset + data.records.length;
+                                this.total = this.offset + this.last.xhr_offset + data.records.length;
```
vitmalina: Are you sure it is correct? This.last.xhr_offset is set before call is made.

unix786: In the test case `this.last.xhr_offset` is less than `this.offset`. I don't know what `this.last.xhr_offset` is supposed to be. I posted further information in the [issue](https://github.com/vitmalina/w2ui/issues/1775#issuecomment-506269214).

vitmalina:
I checked the issue and indeed it is a bug. Thank you for detailed explanation. I am merging this request into main branch.

vitmalina:
@unix786 - thanks for the fix.

PR #1841, opened 2019-06-22
merge

lacmuch:


vitmalina:
The branches diverged too much. I do not think it it is possible to merge it now.

PR #1860, opened 2019-09-12
Fix date file format

StalderT:
A bug was introduce in https://github.com/vitmalina/w2ui/commit/ac2e968e7fc40733c95ee2dc978e2d3bb2e1cf66

Eg: I have :
```
date = "04/03/2020"; // Mar 4 2020
datefomat = "dd/mm/yyyy"; 
```
w2utils.idDate use `new Date(val)` and decode the date as a short dates with "MM/DD/YYYY" format (see https://www.w3schools.com/js/js_date_formats.asp) and decode the date as `Apr 3 2020` insted of `Mar 4 2020`


vitmalina:
Thank you for your contribution. However, `w2utils.isDate` is covered by unit tests. Can you please add unit test for your particular case, you can find it here `qa/index.html`. In my branch currently 3 assertions fail, I am not sure why. Can you see if you can fix it in a way that current test pass and that your new condition is another passing assessment.

StalderT:
Hi @vitmalina 

With my patch all tests pass :

image

I will add my particular case in the unit tests.

StalderT:
I added a test to the PR.

Now all tests pass.

vitmalina:
Awesome. Thank you for your contribution

PR #1864, opened 2019-09-23
optimization w2grid.js

akaJes:
-removed unnecessary event handlers 
-DOM nodes looking more sanitized

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -4733,6 +4733,37 @@
 
             var gridBody = $('#grid_'+ this.name +'_body', obj.box).html(bodyHTML),
                 records = $('#grid_'+ this.name +'_records', obj.box);
+            var frecords = $('#grid_'+ this.name +'_frecords', obj.box);
+            records
+            .on('mouseover', 'tr', function(ev) {
+              $('#grid_'+ this.name +'_frec_' + w2utils.escapeId($(this).attr('recid'))).addClass('w2ui-record-hover')
+            })
+            .on('mouseout', 'tr', function(ev) {
+              $('#grid_'+ this.name +'_frec_' + w2utils.escapeId($(this).attr('recid'))).removeClass('w2ui-record-hover')
+            })
+
+            frecords
+            .on('mouseover', 'tr', function(ev) {
+              $('#grid_'+ this.name +'_rec_' + w2utils.escapeId($(this).attr('recid'))).addClass('w2ui-record-hover')
+            })
+            .on('mouseout', 'tr', function(ev) {
+              $('#grid_'+ this.name +'_rec_' + w2utils.escapeId($(this).attr('recid'))).removeClass('w2ui-record-hover')
+            })
+            var self = this;
+            if(w2utils.isIOS)
+              records.add(frecords)
+              .on('click', 'tr', function(ev) {
+                self.dblClick($(this).attr('recid'), ev);
+              })
+            else
+              records.add(frecords)
+              .on('click', 'tr', function(ev) {
+                self.click($(this).attr('recid'), ev);
+              })
+              .on('contextmenu', 'tr', function(ev) {
+                self.contextMenu($(this).attr('recid'), null, ev);
+              })
+
             // enable scrolling on frozen records,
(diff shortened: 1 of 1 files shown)


src/w2grid.js
```diff
@@ -4733,6 +4733,37 @@
...
+              $('#grid_'+ this.name +'_rec_' + w2utils.escapeId($(this).attr('recid'))).addClass('w2ui-record-hover')
+            })
+            .on('mouseout', 'tr', function(ev) {
+              $('#grid_'+ this.name +'_rec_' + w2utils.escapeId($(this).attr('recid'))).removeClass('w2ui-record-hover')
+            })
+            var self = this;
+            if(w2utils.isIOS)
+              records.add(frecords)
+              .on('click', 'tr', function(ev) {
+                self.dblClick($(this).attr('recid'), ev);
```
vitmalina: this behavior is needed for mobile devices

src/w2grid.js
```diff
@@ -4733,6 +4733,37 @@
 
             var gridBody = $('#grid_'+ this.name +'_body', obj.box).html(bodyHTML),
                 records = $('#grid_'+ this.name +'_records', obj.box);
+            var frecords = $('#grid_'+ this.name +'_frecords', obj.box);
+            records
+            .on('mouseover', 'tr', function(ev) {
+              $('#grid_'+ this.name +'_frec_' + w2utils.escapeId($(this).attr('recid'))).addClass('w2ui-record-hover')
```
vitmalina: Should `this.name` be `self.name` and var self defined above it?

src/w2grid.js
```diff
@@ -7235,20 +7252,6 @@
                     (isRowSelected && this.selectType == 'row' ? ' w2ui-selected' : '') +
                     (record.w2ui && record.w2ui.editable === false ? ' w2ui-no-edit' : '') +
                     (record.w2ui && record.w2ui.expanded === true ? ' w2ui-expanded' : '') + '" ' +
-                (summary !== true ?
```
vitmalina: Did you account for summary records in your changed code?

src/w2grid.js
```diff
@@ -7213,20 +7244,6 @@
...
-                    (w2utils.isIOS ?
-                        '    onclick  = "w2ui[\''+ this.name +'\'].dblClick(jQuery(this).attr(\'recid\'), event);"'
-                        :
-                        '    onclick  = "w2ui[\''+ this.name +'\'].click(jQuery(this).attr(\'recid\'), event);"'+
-                        '    oncontextmenu = "w2ui[\''+ this.name +'\'].contextMenu(jQuery(this).attr(\'recid\'), null, event);"'
-                     )
-                    : ''
-                ) +
-                (this.selectType == 'row' ?
```
vitmalina: Did you account for selectType='cell' in your changed code?

src/w2grid.js
```diff
@@ -4733,6 +4733,37 @@
...
+              })
+            else
+              records.add(frecords)
+              .on('click', 'tr', function(ev) {
+                self.click($(this).attr('recid'), ev);
+              })
+              .on('contextmenu', 'tr', function(ev) {
+                self.contextMenu($(this).attr('recid'), null, ev);
+              })
+
```
vitmalina: Your code does look cleaner, but I am not sure it covers all use cases. Also, please note that due to lazy loading records might not be present, therefore, adding events with `on('click'...` might be wrong. I am not sure, though.

src/w2grid.js
akaJes: You are right, this hadn't been tested well

src/w2grid.js
akaJes: the summary recodrs parent don't have an event handler

src/w2grid.js
akaJes: fixed it in the code above

src/w2grid.js
akaJes: when parent node has an event handler - all children use it
[Event Handler Attachment](https://api.jquery.com/on/)

vitmalina:
I think moving away from inline js is a good idea. Thank you for your contributeion.

PR #1877, opened 2020-01-05
fixed 1876 - options: { autoFormat: false } does not work on columns in the grid"

mdavis1960:
also fixed two javascript console errors on use of obsoleted content()


mdavis1960:
Thanks, I did not know that and I was wondering why there were so many
commits with dist only changes

PR #1889, opened 2020-03-15
Fixed #1878

### LINKED ISSUE #1878: Dates passed to grid in format yyyy-MM-ddTHH:mm:ss are not recognised correctly
* **I'm submitting a ...**
[x] bug report
[ ] feature request

* **What is the current behaviour?**
Dates passed to grid in format yyyy-MM-ddTHH:mm:ss (for example "2020-01-29T18:09:56") are not recognised as dates and in grid written as zero dates (1970-01-01).

* **Other information**
This worked correctly before the commit https://github.com/vitmalina/w2ui/commit/4c58c4baaa7d88cf5e398854590c7d0889057f4a#diff-6ce300965d88ab4e70ced0cd752e9cd8. It is named "Fix date file format". What exactly it fixes?

Vaidot:
Implemented https://github.com/vitmalina/w2ui/issues/1878#issuecomment-597570359 and moved some statements to avoid redundant checks.

### DIFF
--- a/src/w2utils.js
+++ b/src/w2utils.js
@@ -282,19 +282,24 @@ var w2utils = (function ($) {
     }
 
     function isDateTime (val, format, retDate) {
-        if (format == null) format = w2utils.settings.datetimeFormat;
-        var formats = format.split('|');
         if (typeof val.getFullYear === 'function') { // date object
             if (retDate !== true) return true;
             return val;
-        } else if (parseInt(val) === val && parseInt(val) >= 0) {
-            val = new Date(parseInt(val));
-            if (retDate !== true) return true;
-            return val;
-        } else if (parseInt(val) === val && parseInt(val) < 0) {
-            return false;
+        }
+        var intVal = parseInt(val);
+        if (intVal === val) {
+            if (intVal < 0) return false;
+            else if (retDate !== true) return true;
+            else return new Date(intVal);
+        }
+        var tmp = String(val).indexOf(' ');
+        if (tmp < 0) {
+            if (String(val).indexOf('T') < 0 || String(new Date(val)) == 'Invalid Date') return false;
+            else if (retDate !== true) return true;
+            else return new Date(val);
         } else {
-            var tmp = String(val).indexOf(' ');
+            if (format == null) format = w2utils.settings.datetimeFormat;
+            var formats = format.split('|');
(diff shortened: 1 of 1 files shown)


vitmalina:
Can you please explain to me what does the proposed change in code do? If I understand correctly, you change behavior of isDateTime to treat integers as valid dates? If this is the case, then it is incorrect behavior. The function checks if the string is a valid date/time in the format specified.

Vaidot:
I did not change how integer are treated. I simply refactored that part a little to avoid doing check `parseInt(val) === val` twice. I can remove irrelevant changes to avoid confusion.
What i did change is for isDateTime() to treat format yyyy-MM-ddTHH:mm:ss as valid date to correct https://github.com/vitmalina/w2ui/issues/1878 issue.
And i am simply returning feature added in https://github.com/vitmalina/w2ui/pull/1718, but it got removed in https://github.com/vitmalina/w2ui/pull/1860 as a side effect.

Vaidot:
I can explain why this change is necessary in more detail. Grid uses functions w2utils.formatters.date(), w2utils.formatters.datetime() and other similar to output provided values as dates in specified format. Those functions use w2utils.isDateTime() and w2utils.isDate() to convert provided values to date objects. Those date objects then formatted depending on function and provided format.

What values are provided to grid and what you want to be outputted is not necessary the same. If values are given from remote source, then they are likely in JSON format. Tools like JSON.stringify() in JavaScript or [Json.NET](https://www.newtonsoft.com) by default converts dates to yyyy-MM-ddTHH:mm:ss format. So it is convenient to transfer data in standard format and then convert it in grid to a format you want.

Vaidot:
At this rate this pull request is becoming permanent candidate, like others. I would like some sort of response eventually. If you don't like what you see, just reject it, preferably with explanation what you don't like, so I would know what to change to make it acceptable.
If it is unclear what is being changed, I can reduce changes only to relevant to the https://github.com/vitmalina/w2ui/issues/1878 and add some comment that explains why it is being done.
If you you think that searching for a letter 'T' is weird, then I can instead add some grid parameter called tryToConvertDatesUsingNew or somewhat similar. This parameter would be false by default, but if it is set to true, then function would first try to use new Date() to convert string, then other method.
If i get no response i will try reduction of changes method.

PR #1946, opened 2021-02-11
w2field color use rgba for preview

mpf82:
w2field color use rgba for preview.

Before:
color_before

Now:
color_now

Nice to have: a checkered background behind the color preview, so the opacity can actually be perceived

### DIFF
--- a/src/w2fields.js
+++ b/src/w2fields.js
@@ -894,9 +894,13 @@
             if (this.type === 'color') {
                 var color = $(this.el).val();
                 if (color.substr(0, 3).toLowerCase() !== 'rgb') {
-                    color = '#' + color;
-                    var len = $(this.el).val().length
-                    if (len !== 8 && len !== 6 && len !== 3) color = '';
+                    var len = color.length;
+                    if (len !== 8 && len !== 6 && len !== 3) {
+                        color = '';
+                    } else {
+                        var rgb = w2utils.parseColor(color);
+                        color = 'rgba('+ rgb.r +','+ rgb.g +','+ rgb.b +','+ rgb.a +')';
+                    }
                 }
                 $(this.el).next().find('div').css('background-color', color);
                 if ($(this.el).hasClass('has-focus') && $(this.el).data('skipInit') !== true) {


vitmalina:
Strange, modern browsers support #112233aa, 8-char hex colors. Why it did not work there, not sure. Bay be because of leading `#`, possibly we need to check if there is a leading `#` then ignore it?

mpf82:
You are right, that modern browsers should support RGBA in hex notation: https://caniuse.com/?search=rgba%20hex

However, I think that the issue here is with JQuery.

In the code you're setting the color like this: 

``$(this.el).next().find('div').css('background-color', color);``

So if the color was ``#8C96A124``, JQuery turns this value into ``rgb(140, 150, 161)``, losing the alpha channel in the process. This might be also due to the fact that we're still using JQuery 2.1.3

But then again, I cannot reproduce this behaviour with jsfiddle: https://jsfiddle.net/mpnhosgj/

In the fiddle the color is converted correctly to ``rgba(140, 150, 161, 0.14)``.

Feel free to discard this PR if you also cannot reproduce the alpha channel getting lost.

I think for our users, this would be negligible, AFAIK they are rarely using the alpha channel anyway and even then it's working in the application, it was just the preview in the w2field that was off.

Cheers =)

vitmalina:
Got it. I think that if it is jQuery problem, it will be fixed with jQuery. I would keep it as it for now.

mpf82:
If found the problem and fixed it in our code base.

The issue is with JQuery UI that - even in the latest version - comes with jQuery Color Animations v2.1.2 and here the alpha channel is lost when setting Hex8 colors with ``$element.css('background-color', '#11223344');``

I've update our JQuery UI to use jQuery Color Animations v2.2.0 and this fixes the issue.
Unfortunately this is now a custom built JQuery UI with a manual fix, but we're still relying on other JQuery UI functionality, like .draggable, .droppable, an .selectable - otherwise I would have preferred to remove JQuery UI completely from our projects.

I figured this is a valuable information for everyone that also has to include JQuery UI for legacy reasons.

Cheers =)

PR #1950, opened 2021-02-12
new w2overlay align options

mpf82:
added both+left and both+right as w2overlay align options.

with both+X the min width will be the with of the element and the max with is auto, so no more scroll bars for long items

### DIFF
--- a/src/w2utils.js
+++ b/src/w2utils.js
@@ -2656,6 +2656,7 @@ w2utils.event = {
                 var tipLeft  = options.tipLeft;
                 // alignment
                 switch (options.align) {
+                    case 'both+left':
                     case 'both':
                         boxLeft = 17;
                         if (options.width === 0) options.width = w2utils.getSize($(obj), 'width');
@@ -2663,6 +2664,13 @@ w2utils.event = {
                         break;
                     case 'left':
                         boxLeft = 17;
+                        break;
+                    case 'both+right':
+                        if (options.width === 0) options.width = w2utils.getSize($(obj), 'width');
+                        if (options.maxWidth && options.width > options.maxWidth) options.width = options.maxWidth;
+                        var maxWidth = Math.max( w - 10, options.width - 17)
+                        boxLeft = w2utils.getSize($(obj), 'width') - maxWidth;
+                        tipLeft = maxWidth - 30;
                         break;
                     case 'right':
                         boxLeft = w2utils.getSize($(obj), 'width') - w + 10;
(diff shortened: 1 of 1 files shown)


mpf82:
Comparison with one short item:

w2ui_both_left

Comparison with a short and a long item:

w2ui_both_left_2

vitmalina:
I see that you in principle added `both+right`, which just makes tip of the overlay to be at the right side, correct?

mpf82:
The negative tipLeft would be a nice additional feature, but the key difference between left/right and both+left/both+right is, that for "left", the overlay extends to the right of the widget, while for "right" the overlay is moved to the left of the widget.

So it's not just the position of the tip that's different.

Or to put it in other words:
- for "left" the left sides of widget and overlay align
- for "right" the right sides of widget and overlay align

I thought that was your intention when you named the option "align" ;-) 

I've made additional GIFs to show this difference:

Comparison with one short item:
w2ui_both_right_1


Comparison with a short and a long item:
w2ui_both_right_2

vitmalina:
Negative `tipLeft` has been implemented. I am still not sure what it does. I fail to see it in the gifs you added, as they cannot be stopped. Can you add images of `right` and `both+right` to show how it would be different. 

For me alignment means this
- `align=left`, the `left` coordinate of the popup will be aligned with `left` coordinate of the element it is attached to and the `width` of overlay will be based on the content of overlay. It can be smaller than `width` of the element or greater.
- `align=right`, the `right` coordinate of the popup will be aligned with `right` coordinate of the element it is attached to and the `width` of overlay will be based on the content of overlay. It can be smaller than `width` of the element or greater.
- `align=both`, the `left` and `right` coordinate of the popup will be aligned with `left` and `right` coordinate of the element it is attached to and the `width` of overlay will be same as `width` of the element, regardless content in the overlay (**this case is best for dropdown controls**)

mpf82:
``both+left`` and ``both+right``  behave like ``both`` for items that are smaller than the width of element it's attached to.

``both+left`` behaves like ``left`` if the items are wider than the width of element it's attached to.

respectively ``both+right`` behaves like ``right`` if the items are wider than the width of element it's attached to.

Our use case is exactly this - the overlay width should be **minimum** the width of the input, but instead of showing scrollbars for large items, the overlay should grow accordingly, either to the left or the right.

If that explanation was not clear enough, let me know and I will add still images to clarify.

vitmalina:
What about `minWidth` = [`auto`, `input`, `XXpx`], with `auto` being default for all and `input` being default for `align=both`

mpf82:
> 
> 
> What about `minWidth` = [`auto`, `input`, `XXpx`], with `auto` being default for all and `input` being default for `align=both`

Sounds good.

I will redesign my PR in the next few days.

mpf82:
Closing this PR in favor of #1962

PR #1954, opened 2021-02-19
Fix w2grid getCellHTML title

mpf82:
Make title behave consistently accross the various renderers.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -7402,6 +7402,7 @@
         },
 
         getCellHTML: function (ind, col_ind, summary, col_span) {
+            var obj = this;
             var col = this.columns[col_ind];
             if (col == null) return '';
             var record        = (summary !== true ? this.records[ind] : this.summary[ind]);
@@ -7478,9 +7479,21 @@
                     '></span>';
             }
             // various renderers
+            function getTitle(cellData){
+                var title = "";
+                if (obj.show.recordTitles) {
+                    if (col.title != null) {
+                        if (typeof col.title == 'function') title = col.title.call(obj, record, ind, col_ind);
+                        if (typeof col.title == 'string')   title = col.title;
+                    } else {
+                        title = w2utils.stripTags(String(cellData).replace(/"/g, "''"));
+                    }
+                }
+                return (title != null) ? String(title) : "";
+            }
             if (col.render != null && ind !== -1) {
                 if (typeof col.render == 'function') {
-                    var html = col.render.call(this, record, ind, col_ind, data)
+                    var html = col.render.call(this, record, ind, col_ind, data);
                     if (html != null && typeof html == 'object') {
                         data = $.trim(html.html || '');
                         addClass = html.class || '';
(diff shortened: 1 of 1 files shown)


mpf82:
In case someone wants to write a unit test:

```
<div id="grid" style="height: 300px;"></div>

function test_grid() {
    var render_obj                = {
        "Hannibal the ..." : "Hannibal the ... object",
    };
    var render_func               = ( record, ind, col_ind, data ) => data + " func";
    w2utils.formatters.render_str = ( data, params, record ) => data + " str";
    var render_title              = data => "My custom title";
    $().w2destroy( "grid" );
    $( "#grid" ).w2grid( {
        name    : "grid",
        columns : [
            { field : "name", caption : "render:null", render : null },
            { field : "name", caption : "render:obj", render : render_obj },
            { field : "name", caption : "render:func", render : render_func },
            { field : "name", caption : "render:str", render : "render_str" },
            // ----
            { field : "name", caption : "render:null NT", render : null, title : "" },
            { field : "name", caption : "render:obj NT", render : render_obj, title : "" },
            { field : "name", caption : "render:func NT", render : render_func, title : "" },
            { field : "name", caption : "render:str NT", render : "render_str", title : "" },
            // ----
            { field : "name", caption : "render:null TF", render : null, title : render_title },
            { field : "name", caption : "render:obj TF", render : render_obj, title : render_title },
            { field : "name", caption : "render:func TF", render : render_func, title : render_title },
            { field : "name", caption : "render:str TF", render : "render_str", title : render_title },
        ],
        records : [
            {
                recid : 1,
                name  : "Hannibal the ...",
            },
        ],
        total : 1,
    } );
}
test_grid();
```

src/w2grid.js
```diff
@@ -7478,9 +7479,21 @@
                     '></span>';
             }
             // various renderers
+            function getTitle(cellData){
```
vitmalina, line 7482: The only suggestion is to move this function to the end of the parent function. I think it is more readble to have nested functions at the end. Even after `return`. I will do it.

src/w2grid.js
```diff
@@ -7489,14 +7502,15 @@
...
-                        data = '<div style="'+ style +'">' + infoBubble + String(data) + '</div>';
+                        data = '<div style="'+ style +'" title="'+ getTitle(data) +'">' + infoBubble + String(data) + '</div>';
                     }
                 }
                 // if it is an object
                 if (typeof col.render == 'object') {
                     var dsp = col.render[data];
                     if (dsp == null || dsp === '') dsp = data;
-                    data = '<div style="'+ style +'">' + infoBubble + dsp + '</div>';
+                    data = dsp;
```
vitmalina, line 7512: this line is confusing, as in the next line data is set again

vitmalina (approved):

src/w2grid.js
mpf82: true, I only did that so the line ``data = '<div style="'+ style +'" title="'+ getTitle(data) +'">' + infoBubble + String(data) + '</div>';``  is the same for all cases

Issue #708, opened 2014-11-07
w2grid Columns Re-order probrem


istoshi:
I've found a problem in reorder grid column drag and drop without mousemove event,
column position moved to first position.
it caused by "_dragData.lastInt" is initialized by null and not set.

it occured in this controll:
http://w2ui.com/web/demos/#!grid/grid-25

BangisBanginis:
I second that. 
I can recreate simmilar behavior with latest (v1.5) w2ui.grid component.
1. Used sample code from demos (http://w2ui.com/web/demos/#!grid/grid-25). 
2. Made columns sortable.
3. running test page.
4. Pushing mouse down while over column header and NOT moving the mouse but HOLDING it even fore a fraction of a second. 
5. When releasing - column jumps into 1st position.

omiinaya:
this is still an issue kind of. you don't need to hold mouse to drag, meaning that you can't make them re-orderable and sortable at the same time because clicking at all starts the dragging process.

TauerOne:
Here is a solution to this issue.  In the initColumnDrag function, initialize _dragData.lastInt to undefined instead of null.  

The difference is that when null is used in Math.min(), it is coerced to a Number and treated as zero, similar to the way isNaN(null) returns false.  Math.min(null, 10) will return 0, while Math.min(undefined, 10) will return NaN.  

That comes into play around the 16th line of the dragColEnd function.  If _dragData.lastInt is undefined then targetColumn will be an empty jQuery object with a length property of 0, which will prevent the dragColEnd function from moving the column.  

If you actually drag the column and cause the blue column drag marker to appear, then _draData.lastInt will be set to the number of the column you are dragging to, and everything will work as it should.

mpf82:
Closing as fixed

PR #1992, opened 2021-04-02
Fix Columns Re-order problem #708

TauerOne:
In the initColumnDrag function, initialize _dragData.lastInt to undefined instead of null.

The difference is that when null is used in Math.min(), it is coerced to a Number and treated as zero, similar to the way isNaN(null) returns false. Math.min(null, 10) will return 0, while Math.min(undefined, 10) will return NaN.

That comes into play around the 16th line of the dragColEnd function. If _dragData.lastInt is undefined then targetColumn will be an empty jQuery object with a length property of 0, which will prevent the dragColEnd function from moving the column.

If you actually drag the column and cause the blue column drag marker to appear, then _draData.lastInt will be set to the number of the column you are dragging to, and everything will work as it should.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -5428,7 +5428,7 @@
 
             var obj = this,
                 _dragData = {};
-                _dragData.lastInt = null;
+                _dragData.lastInt = undefined;
                 _dragData.pressed = false;
                 _dragData.timeout = null;_dragData.columnHead = null;
 


TauerOne:
P.S.  I made and tested this change in 1.5-rc1, since column re-ordering seems to be disabled or not working in the latest 1.5 branch.

PR #1998, opened 2021-04-16
jQuery.isArray is deprecated

mpf82:
https://api.jquery.com/jQuery.isArray/

Note: This API has been deprecated in jQuery 3.2; please use the native Array.isArray method instead.

### DIFF
--- a/src/w2fields.js
+++ b/src/w2fields.js
@@ -395,7 +395,7 @@
                     options.items    = w2obj.field.prototype.normMenu.call(this, options.items);
                     options.selected = w2obj.field.prototype.normMenu.call(this, options.selected);
                     this.options = options;
-                    if (!$.isArray(options.selected)) options.selected = [];
+                    if (!Array.isArray(options.selected)) options.selected = [];
                     $(this.el).data('selected', options.selected);
                     if (options.url) {
                         options.items = [];
@@ -430,7 +430,7 @@
                     };
                     options = $.extend({}, defaults, options);
                     this.options = options;
-                    if (!$.isArray(options.selected)) options.selected = [];
+                    if (!Array.isArray(options.selected)) options.selected = [];
                     $(this.el).data('selected', options.selected);
                     if ($(this.el).attr('placeholder') == null) {
                         $(this.el).attr('placeholder', w2utils.lang('Attach files by dragging and dropping or Click to Select'));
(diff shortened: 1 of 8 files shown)


vitmalina:
MIke, I have already done this change, but did not put it. I did not know it was deprecated, thought. Starting with what version?

vitmalina:
It might be a good idea to submit this PR into `w2ui-1.5` branch? Can you open PR to it. Also, I have been working with a friend to convert w2ui into ES6 modules. I really like how clean it becomes. It is published in es6 branch. ($.isArray is fixed there). However, it will not merge cleanly as every line changed. I want to know what you think. Can you please look at ES6 branch in /src folder. The API did not change and every call is the same. It does change the way objects are created, but I think we can make some sort fo sham for 1.5 compatibility. ES6 modules will only work in evergreen browsers, but it could be babeled for older browser support if needed.

mpf82:
Hi Vitali,

> I did not know it was deprecated, thought. Starting with what version?

It's deprecated since JQuery 3.2 - so since March 2017 ( See: https://en.wikipedia.org/wiki/JQuery#Release_history )

> It might be a good idea to submit this PR into w2ui-1.5 branch?

I don't think so. I was planing to replace most of the deprectated JQuery functions - at least those that can be replaced with vanilla JS -  step by step (see https://api.jquery.com/category/deprecated/ - so $.trim and $.isFunction, ... would have been next), but as I understand, everything that's flaged as deprecated in JQuery 3.x will still stay throughout all coming 3.x releases and will only be removed in a yet-to-be-announced [JQuery 4.0](https://github.com/jquery/jquery/milestone/7) - so I think there's no reason to bring these changes to the 1.5 branch.

Or do you think differently? Should the 1.5 branch still receive non-critical updates?

> However, it will not merge cleanly as every line changed. I want to know what you think. Can you please look at ES6 branch in /src folder.

Ha, it seems you've made use of the ESLint auto-fix feature and since you're using ``"semi": ["error", "never"]`` it's no wonder every line changed.

I agree, it's looking cleaner that way.

> ES6 modules will only work in evergreen browsers, but it could be babeled for older browser support if needed.

Also agreed. 

---

Which branch do you want to focus on for the 2.0 release?

The current master branch or are you planing to make the es6 branch the new master in the near future?

Off-Topic: it seems you did not yet reply to our E-Mail from March, 1st

Cheers, Mike

vitmalina:
ES6 will become Master soon.

mpf82:
Closing this PR, as it's already fixed in the es6 branch and the es6 branch will become master soon.

PR #1999, opened 2021-04-19
Fixed errors in setTimeout due to hoisting

unix786:
There was a problem with variable scope. Fixed it by replacing `var` with `let`.

The problem seems to have been caused by hoisting within the scope of function `uncheck`: each iteration of loop `for` was actually using the same variable, so after calling `setTimeout`, the value sometimes changes and sometimes also becomes `null`, which would result in errors.

The same problem exists in master branch, so I'll create another pull request for that. Brach "es6" does not have this problem since all `var` expressions were changed into `let`.

### DIFF
--- a/src/w2toolbar.js
+++ b/src/w2toolbar.js
@@ -303,7 +303,7 @@
             var items = 0;
             var tmp   = [];
             for (var a = 0; a < arguments.length; a++) {
-                var it = this.get(arguments[a]);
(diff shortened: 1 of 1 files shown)


vitmalina:
This is already fixed in ES6 branch which will soon be merged to master. There is lots of clean up, including all `var` are replaced with `let`

PR #2000, opened 2021-04-19
Fixed errors in setTimeout due to hoisting

unix786:
Uncheck function was producing errors due to hoisting of `var it`. After calling `setTimout`, the `for` loop may continue and value of `it` changes and may even become `null`.

The same fix was done for branch w2ui-1.5 in commit unix786/w2ui@220c4a73 (pull request #1999 ).


vitmalina:
This is already fixed in ES6 branch which will soon be merged to master. There is lots of clean up, including all var are replaced with let

PR #2025, opened 2021-05-27
updated w2form docs and demos, improved w2form.applyFocus, Recompiled JS

mpf82:
updated w2form docs and demos, improved w2form.applyFocus, Recompiled JS


vitmalina:
Also, please do separate PR for /dist/* file. Often it creates conflicts as I might have been building it locally, and if it is a separate request, it is easier to merge or ignore.

Issue #2024, opened 2021-05-27
Grid gets stuck with inEditMode flag set


ECrosland:
**Short description**
I noticed that my grid would enter a state where double clicking an editable cell would no longer enter into editing mode on that cell. Such edits had been working just moments earlier in the same browser window.

**Steps to reproduce or sample** 
This situation comes about when I try to arrange for edits to a cell to be submitted immediately back to the server. I do this by placing a call to the grid's `save` method within an event handler for the grid's `change` event.

It seems that what happens is that the save request is sent to the server, and then the selection moves to the next cell in the column. So far so good.

When the save request completes, the grid then refreshes itself. As part of this, a `refreshBody` method is called, which replaces the HTML for the grid body, and we thereby lose the `input` element that had been present for editing. However, although this element goes away, `grid.last.inEditMode` remains set to true. When, therefore, `editField` is called again in response to a double click, after checking this flag, it appears to the implementation as though editing is already underway, and it does not create a new `input` element for editing whichever cell was just double clicked.

This, at least, is what my time spent with the debugger seemed to reveal to me.

It seems to me there's a bug here, although I know it's also possible I'm misusing the API in some way.

As a side note, I'd be happy to keep the edit cursor from automatically advancing to the next cell in the column, since that doesn't really suit my use case anyway. That should have the happy side effect of keeping this problem from manifesting for me.

ECrosland:
I was wondering why I didn't end up in this situation if I invoked `save` in response to a button click.

It seems that the `blur` handler on line 2956 in w2grid.js (and in comparable places on other switch branches) is ordinarily used to remove the `input` element and reset the internal state, including the `inEditMode` flag. However, this handler isn't invoked when the body HTML is replaced, as it is in the sequence of events described above.

https://github.com/vitmalina/w2ui/blob/4d98cee769d1df08686e4afc9b35bb7eff7b3c28/src/w2grid.js#L2956-L2959

vitmalina:
If PR mentioned here resolve provide a solution, please close this issue.

ECrosland:
The PR that adds `editAutoAdvance` allows me to avoid ending up in the situation where the identified problem manifests, and this is fine in my use case, since I would prefer not to have the auto-advance behavior anyway.

[...]

PR #2028, opened 2021-05-27
Make it possible to disable automatic cursor advance after making an inline edit

ECrosland:
I mentioned as a side note in bug report #2024 that I would like to prevent the behavior whereby, when enter is pressed to submit an inline edit, the "edit cursor" automatically advances to another editable cell and opens an inline editor for that cell.

I couldn't find an option whereby I might disable this behavior, so I attempted to introduce one. I've called the option `advanceOnEdit`, and it is set to `true` by default to preserve existing behavior. My code changes check this setting in two places; I'm not sure whether I've caught all necessary places where the check should be made, so please review.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -262,6 +262,7 @@ class w2grid extends w2event {
         this.method            = null // if defined, then overwrites ajax method
         this.dataType          = null // if defined, then overwrites w2utils.settings.dataType
         this.parser            = null
+        this.advanceOnEdit     = true // automatically begin editing the next cell after submitting an inline edit?
 
         // these column properties will be saved in stateSave()
         this.stateColProps = {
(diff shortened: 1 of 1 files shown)


ECrosland:
FWIW, I have a workaround that seems to prevent this "auto-advance" behavior in a recent build of w2ui. It basically checks whether the event that leads to a w2ui `editField` event was triggered on a `w2ui-input` element. This feels a bit less than clear, but it does seem to successfully distinguish the case where `editField` is called on a subsequent cell in response to Enter being pressed in the editor for a previous one. Then I can simply cancel the event.

```javascript
myGrid.on("editField", function (event) {
	// NB: Seems to only work when selectType is "row", which is OK for my needs.
	const isAutoAdvance = $(event.originalEvent.target).hasClass("w2ui-input");
	if(isAutoAdvance) {
		event.preventDefault();
	}
});
```

vitmalina (approved):

vitmalina:
@ECrosland thank you for you contribution. I think it is a good feature. The only thing I am not sure `advanceOnEdit` is the best name. What about `editAtuoAdvance` - I like that it starts with `edit` as it will be grouped with similar properties/methods, such as `editField`.

Also, I noticed you pushed to master. Are you already using version of master in your dev? Nice, if it is so. Thought master is still buggy at times. But we are working on it.

ECrosland:
@vitmalina Your thinking on the name makes sense to me.

As for my pushing to master--actually, it's just that I didn't really think it through. I'm still using 1.5 in my project, not 2.0, so, I guess it would be helpful to me to have the change on the 1.5 branch also.

ECrosland:
Just created #2032 which adds the option on branch w2ui-1.5. This is the branch that the copy of w2ui in my latest project is built from.

PR #2034, opened 2021-05-31
updated w2ui-1.5* files in dist folder (master)

mpf82:
updated w2ui-1.5* files in dist folder (master)


vitmalina:
@mpf82 I think this is unnecessary. I think in 1.5 branch we should keep nightly version under `w2ui.js`, not `w2ui-1.5.js` and when there are some fixes, we can release it as 1.5.1. No need to change w2ui-1.5.js as it would be confusing. I will retract previous PR.

PR #2077, opened 2021-07-23
w2popup.js

mansurdegirmenci:
to center the header.

### DIFF
--- a/src/w2popup.js
+++ b/src/w2popup.js
@@ -156,7 +156,7 @@ class w2dialog extends w2event {
                     if (tmp.length > 0) options.body = tmp.html(); else options.body = parts.html()
                 }
                 // then content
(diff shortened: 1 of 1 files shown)


mpf82:
The header **was** centered in the "old style". It has been commented out in the less file: https://github.com/vitmalina/w2ui/blob/master/src/less/src/popup.less#L43

You can still see the old styling in the demos: http://w2ui.com/web/demos/#/popup/1

New styling: http://jsfiddle.net/75wzagnx/

If you want the title centered, simply add

    .w2ui-popup-title {
    text-align: center;
    }

to yor project specific CSS file.

Fiddle: http://jsfiddle.net/75wzagnx/1/

Side note: on the road to w2ui 2.0 we're trying to get rid of styling instructions in the generated HTML code and move it to the CSS file instead (work in progress).

Issue #1715, opened 2017-11-15
opposite of "w2utils.escapeId"


rolfn:
* **I'm submitting a ...**
[ ] bug report
[ x] feature request
[ ] question about the decisions made in the repository

Sometimes it would be helpful to have the inverse function to "w2utils.escapeId" (--> "unescapeId"). I use path-like names which must be escaped in normal usage but un-escaped if the path should use otherwise. I defined by myself the following:
```
function unescapeId(id) {
  return String(id).replace(/\\/g, '');
}
``` 
Thanks in advance.

mpf82:
Closing as implemented: #2105

PR #2105, opened 2021-09-22
w2utils.unescapeId

mpf82:
Added unescapeId() to w2utils, as requested in #1715

Since w2ui uses JQuery anyway, I've changed w2utils.escapeId() to call $.escapeSelector() instead.

You can test both methods like this:

    const s = String.raw`/([;&,\.\+\*\~'\`:"\!\^#$%@\[\]\(\)=<>\|\/? {}\\])/`
    w2utils.unescapeId( w2utils.escapeId(s) ) === s

### DIFF
--- a/src/w2utils.js
+++ b/src/w2utils.js
@@ -15,6 +15,7 @@
 *   - normMenu
 *   - w2utils.message - return a promise
 *   - bindEvents - common method to avoid inline events
+*   - unescapeId
 *
 ************************************************/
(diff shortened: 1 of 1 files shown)


mpf82:
I admit that ``$.find.selectors.preFilter.ATTR([null, id])[1]`` does not look sexy, but unfortunately, JQuery does not expose Sizzle's ``runescape`` and ``funescape`` directly.

PR #2106, opened 2021-09-22
Fixed name checking

mpf82:
The check ``if (name)`` would allow grids (and other w2ui objects) with the names ``false, 0, null, undefined, ""`` to be created without an error message and without being registered to the global ``w2ui`` object.

### DIFF
--- a/src/w2event.js
+++ b/src/w2event.js
@@ -10,7 +10,7 @@ class w2event {
     constructor(name) {
         this.handlers = []
         // register globally
-        if (name) {
+        if (typeof name !== 'undefined') {
             window.w2ui = window.w2ui || {}
             if (!w2utils.checkName(name)) return
             window.w2ui[name] = this


vitmalina:
Actually, not true. `false, 0, null, undefined, ""` are not allowed by `if (name)...` In addition there is a `checkName` call that would require alpha numeric names. I think we could keep it as is, or 
```
if (typeof name == 'string') ...
```

mpf82:
> In addition there is a checkName call that would require alpha numeric names.

But we're not running into ``checkName``, because we're not entering the ``if`` block, if the name is ``false, 0, null, undefined, ""``

http://jsfiddle.net/d5w78fue/

The idea is to always enter the if block (and thereby checkName) if the name is anything but ``undefined``

vitmalina:
I see. Well, `w2field` object extends `w2event` and fields do not have names. And they should not be registered in `w2ui` global object. I think that condition was added for it. Then `w2field` need to be modified.

vitmalina:
And if user wants to create a grid with name `null`, let him. JS allows it ))

mpf82:
Ok =)

Closing PR.

vitmalina:
On the second thought, you are right. It will not mess `w2field`. It only triggers when `name` is something. Ok, lets do it

Issue #1991, opened 2021-04-02
Support accent in localSort() for grid


antoninmeunier:
**Describe the problem your feature will solve.**
Sorting the grid with the sort() or localSort() method doesn't handle the accents, for exemple in French : É is sorted after Z, which is not true. 

**Describe the solution you'd like**
Sorting methods should consider accented characters and non-accented characters as the same. 

**Describe alternatives you've considered**
I bypassed this method and created my own method directly in JS file but I guess it should be included in the source code.

vitmalina:
submit a pull request

mpf82:
I think that could easily be done by passing ``Intl.Collator().compare`` as sortMode. Docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/compare

It's supported by all modern browsers: https://caniuse.com/?search=Intl.Collator.compare

We could also add it as an internal feature, as addition to "natural", e.g. by passing ``international`` as sortMode.

Comparison:

```
['1', '10', '2', '20', '3', 'a', 'ä', 'e', 'é', 'i', 'o', 'ö', 'u', 'ú', 'ü',].sort()
['1', '10', '2', '20', '3', 'a', 'e', 'i', 'o', 'u', 'ä', 'é', 'ö', 'ú', 'ü']

['1', '10', '2', '20', '3', 'a', 'ä', 'e', 'é', 'i', 'o', 'ö', 'u', 'ú', 'ü',].sort(w2utils.naturalCompare)
['1', '2', '3', '10', '20', 'a', 'e', 'i', 'o', 'u', 'ä', 'é', 'ö', 'ú', 'ü']

['1', '10', '2', '20', '3', 'a', 'ä', 'e', 'é', 'i', 'o', 'ö', 'u', 'ú', 'ü',].sort(Intl.Collator().compare)
['1', '10', '2', '20', '3', 'a', 'ä', 'e', 'é', 'i', 'o', 'ö', 'u', 'ú', 'ü']
```

However, combining natural sort **and** international sort would be a whole different story.

mpf82:
Closing issue: implemented in PR #2111

PR #2111, opened 2021-09-27
w2utils.internationalCompare

mpf82:
w2utils.internationalCompare / language-sensitive string comparison

Feature request: #1991

Also updated jsFillde API url from http://jsfiddle.net/api/post/jquery/2.1/ to http://jsfiddle.net/api/post/jquery/3.4/

### DIFF
--- a/demos/examples/utils/11.html
+++ b/demos/examples/utils/11.html
@@ -1,7 +1,7 @@
 <div class="content">
     <div id="example_title">
         <h1>International Compare</h1>
-        There is a function, w2utils.internationalCompare, that can be used for language-sensitive sorting of arrays.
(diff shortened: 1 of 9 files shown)


mpf82:
> 
> 
> Good solid work. I am glad to see all aspects, such as docs/demos with the implementation. The only thing, I would propose to rename `internationalCompare` to `i18nCompare`, it is shorter and often abbreviated like it. Both in w2utils and w2grid.

Sure. Should the ``sortMode`` also be changed to ``'i18n'`` or should it stay ``'international'``?

vitmalina:
yes, change sortMode too. `i18n` is easier then to spell the whole word )

mpf82:
@vitmalina just a heads-up that the names have been changed. in case you didn't get a notfication.

PR #2123, opened 2021-10-07
Trying again - adding support for PgUp, PgDn, Home & End keys :)

H-Lo:
Tried to keep up all the formatting as it was before. Also now chages are made into `src` directory, not `dist`. Hope this one will be better. Please check and let me know is there anything else to fix - no problem at all.

Summary:

moved all code from switch command case 38: to separate function moveUp() and made a call to it
moved all code from switch command case 40: to separate function moveDown() and made a call to it
modified prevRow() and nextRow() by ading 3rd argument numRows and adding some logic to support it
updated calls to prevRow() and nextRow() by adding 3rd argument
added new items to switch command to support additional keys: case 33: for PgUp, case 34: for PgDn, case 35: for End and case 36: for Home


vitmalina:
As far as setting up `gulp` is concerned, it is quite easy to set it up. See https://github.com/vitmalina/w2ui#building, once you installed `node` (and `npm`), you just run `npm install` and it will install all dependencies. Then just run `gulp` command from command line and it will recompile js and css files.

vitmalina:
I implemented page soze

H-Lo:
Hi! I would like to check with you @vitmalina  before opening a ticket - it looks like autoload is not working anymore.
My firt thougt was I made a bug in my code and it stopped working after merge, but then I checked out previous version from 2021-09-26 and it's still not working. When I swhitch to 1.5 it works as expected.
I use Java for backend and POST method for data fetch.
If you need more data, I can provide anything needed.

H-Lo:
> Hi! I would like to check with you @vitmalina before opening a ticket - it looks like autoload is not working anymore. My firt thougt was I made a bug in my code and it stopped working after merge, but then I checked out previous version from 2021-09-26 and it's still not working. When I swhitch to 1.5 it works as expected. I use Java for backend and POST method for data fetch. If you need more data, I can provide anything needed.

Oh my.... I found what's going on. I changed font size by CSS and it stopped working after that change. 1.5 works fine because CSS classes are not the same and 1.5 is not affected by those font size changes.
Ok, so it's not a bug, but it's not possible to change font size. I tried to do this because on 4k monitors it looks pretty small.

PR #2132, opened 2021-10-16
issues 2053

mansurdegirmenci:
I will remove the other signs '>', '<', '>=', '<=' and replace them with text software. some time to learn the library. Quite difficult as a "back-end" person :P


mansurdegirmenci:
of course not mandatory. an optional change. Now, when you type and search without opening "search-popup", it already sends "int" fields as "equal". When I sent it with "search-popup" it was sending "=". I wanted to fix it. Also, when it comes to translation, it is necessary to translate "between". expressions(=,>,<) must be symbols or text(between,equal,graterthan). I advocate this. thanks.

PR #2146, opened 2021-11-19
added noRefresh param to expand/collapse

mpf82:
when doing a lot of collapsing/expanding the constant refresh is a performance killer

It's more performant to call collapse/expand with ``noRefresh`` and then call ``grid.refresh()`` once when done.


vitmalina:
Please also update docs, as they are current.

mpf82:
> Please also update docs, as they are current.

Done #2149

PR #2150, opened 2021-11-22
Updated Advanced Row/Cell Styling

mpf82:
Show how to use w2ui.style.column and w2ui.class.column


vitmalina:
Would it be better not to do it on a timeout but on a button click?

mpf82:
> Would it be better not to do it on a timeout but on a button click?

Agreed. Will change when I have the time.

PR #2152, opened 2021-11-23
improved grid styling demo

mpf82:
improved grid styling demo


vitmalina (approved):

vitmalina:
Like it, but would recommend to do this changes
```js
function style_toggle(recid) {
    const ind = w2ui.grid.get(recid, true)
    const record = w2ui.grid.records[ind]
    ...
    w2ui.grid.refreshRow(recid, ind)
}
```

Issue #1964, opened 2021-02-27
BUG in w2grid select & scroll with jquery-3.5.1


Rimokas:
**Short description**
The latest w2ui, downloaded at 21 of February. W2grid infinite vertical scrolling and selection isn't working true with latest directed jquery-3.5.1. 

**Steps to reproduce or sample** 

Let say, that you have a grid with some pages, to use with scrolling to bottom.  
- Select first cell in the grid:

![image](https://user-images.githubusercontent.com/12154604/109387621-2195d200-790b-11eb-836f-641d90ae5f79.png)

- then scroll some to bottom and then return to the top : 

![image](https://user-images.githubusercontent.com/12154604/109387693-979a3900-790b-11eb-9581-f08fd0b1bfde.png)

The select cell remain in the same place, but frame of selection appears rather below. 

 - If to go to bottom and select the last record : 

![image](https://user-images.githubusercontent.com/12154604/109387848-548c9580-790c-11eb-9c01-383b4fbe5207.png)

differs selection field with cell in one row and scroll is the middle, not in bottom ... 

With jquery-3.1.1 all is working fine. 

...

mpf82:
Possibly a duplicate of #1890

Issue #1890, opened 2020-03-16
Grid with selectType = Cell and scroll


ponyEdp:
Good morning
If I set selectType = Cell in a grid.
After I scroll down the grid and I want type a text in a cell on the row number 7, for example, , automatically the grid focus go to the last row and the text (I'm typing) is inserted in the last row.
Watch the animated gif attached.
May you fix the problem ?
Do you want a JsFiddle example or I must post here the program code ?
Thanks
Best regards
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/55495773/76735274-db307d00-6764-11ea-8627-cd504728bf3e.gif)

vitmalina:
Can you please create a JS fiddle example as I am not following what the problem is?

ponyEdp:
Good morning.
I solve the problem.
It was because I use jquery version 3.3.1 instead of jquery version 2.1.1
Regards

shayolz:
Hello, we are having the same issue with selectType cell.

If we scroll on table and try to select something, the selected cell is shifted, like @ponyEdp showed.

This issue appear using Jquery > 3.2. @vitmalina Do you think is there any possibility to fix the problem on last version of jquery? We can't downgrade jquery atm.

You can easly test the issue here:
http://jsfiddle.net/skce016d/

HOW TO reproduce the error:

Scroll a little bit the table, then do a big select of 2-3 records and you will see that the selection is wrong. (While if you select jquery 3.2 there is no bug.)

vitmalina:
It has been fixed in both branches, `master` and `w2ui-1.5`

mpf82:
@vitmalina The issue still exists in the current master, as demonstrated in the linked demo above: http://jsfiddle.net/skce016d/

How to reproduce: scroll down a bit and the make a selection across all 3 columns.

![w2ui_issue_1890](https://user-images.githubusercontent.com/10152404/118566586-71e73800-b774-11eb-9ed2-42665fc80414.gif)

mpf82:
Should be fixed, please re-open if the issue still exists.

PR #2154, opened 2021-11-24
Grid fixes

mpf82:
Removed CSS height from ``.w2ui-col-number`` - height should be determined by recordHeight which is set on the parent TR.

Exit early in ``mouseMove()`` if event.target has no tagName attribute

Fixed Excel-like grid selection

CAUTION: I do not know why the existing code was so complex. Maybe it was due to a different JQuery version? That's why the old code was commented out and not removed!

Please review these changes carefully!

Related issues: #1890 and #1964

### DIFF
--- a/demos/examples/combo/3.html
+++ b/demos/examples/combo/3.html
@@ -18,9 +18,9 @@
 <div id="main" style="width: 100%; height: 300px; margin-bottom: 10px;"></div>
 <button class="w2ui-btn" onclick="mark()">Mark Range C3:D7</button>
 <button class="w2ui-btn" onclick="w2ui.grid.removeRange('range')">Remove Range</button>
-&nbsp;&nbsp;&nbsp;
 <button class="w2ui-btn" onclick="cellFormat(true)">Cell Format</button>
 <button class="w2ui-btn" onclick="cellFormat(false)">Remove Cell Format</button>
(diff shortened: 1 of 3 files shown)


mpf82:
Old selection behavior when scrollig both vertically and horizontally:

grafik

New selection behavior:

grafik

Still not happy with the 1px-off if using frozen columns:

grafik


But no idea how to fix :/

mpf82:
Additional note: the selection will go totally nuts if a column from the "middle" of the grid is frozen (e.g. only freezing column 'C')

I will not even try to fix this - but it seems noboby did notice until now, so let's not mention this ever again ;)

vitmalina:
I have a project where spreadsheet grid is used significantly and mostly it is working. So, if a bug was not found, likely, it is not very visible. Also, just a note, there is colspan, which might effect this rendering. Possibly need to check it

mpf82:
@vitmalina  Do you know what JQuery version your project is using?

The demos page currently uses jquery 3.5.1

Can you check and confirm that currently the spreadsheet demo http://w2ui.com/web/demos/#/combo/3 has a "bad" selection range when the grid is scrolled? I can see the range misbehaving in both Firefox and Chrome.

This PR seems to fix this, but there certainly way a reason for the previous code, that is now commented.

PR #2159, opened 2021-11-30
w2utils.locale accept array

mpf82:
w2utils.locale now accepts an array as the "locale" argument. if "locale" is an array, w2utils.locale will be called recursively for each array item.

also: added German translation for demos

also: demos index now tries to load the locales from demos/locale/ AND also from https://rawgit.com/vitmalina/w2ui/master/src/locale/

I am new to using Promises, so please feel free to correct the code if needed.


vitmalina:
Also, please update documentation for w2utils.locale

vitmalina:
@mpf82 what do you think about removing callBack at all. It already returns a promise?

mpf82:
Like I said, I am not very familiar with the concept of promises and I didn't want to break compatibility.

vitmalina:
Well, promises are awesome and a new way to do async things in JS. w2utils.locale already supports promises and I try to add it every where where there are async operations. Often, I leave callbacks in tack, exepcially if they are in an options object. However, when it is as a param, then it make js not very pretty, here is an example
```js
w2utils.locale(file, () => {
    // refresh once file is loaded
}, true) 
```
Please note, that `true` at the end can easily be missed because it is after the call back.

Same with promise would be

```js
w2utils.locale(file, true).then(() => {
  // refresh once file is loaded
})
```

Let me merge this request and then I will update removing call back. Thanks.

PR #2217, opened 2022-05-19
Update 3.html

squadjot:
Small correction to make the expand example work


squadjot:
There's a small visual bug aswell, when collapsing the row, the padding of the expanded content makes the animation jumpy when the row collapses again.

I'm able to hack a fix with css

```
.w2ui-expanded-row > td > div {
    padding: 0px !important; // far from optimal
}
```

PR #2239, opened 2022-06-01
Add ukrainian locale to demo

AlexKovynev:


vitmalina:
Please note that all translation phrases need to be added ad `demos/locale` in order for them to work with demos.

Issue #2237, opened 2022-05-29
Wrong parse of 'dd.mm.yyyy|h24:mm'


AlexKovynev:
https://w2ui.com/web/demos/#/fields/2
![1](https://user-images.githubusercontent.com/3380440/170865147-0f160ab9-988c-4a6b-9ce7-1834d5874144.png)

It look like 'dd/mm/yyyy|h24:mm' parsed wrong in datetime mode

"mm/dd/yyyy|hh24:mm" works correctly but "dd/mm/yyyy|hh24:mm" is not

PR #2240, opened 2022-06-02
fix datetime tooltip

AlexKovynev:
Look like work after it. Maybe is not all but bug with date fixed.

fix https://github.com/vitmalina/w2ui/issues/2237


vitmalina (commented):
Line #2365 should have also be changed. I will do it.

PR #2244, opened 2022-06-05
Fix ukrainian

AlexKovynev:


AlexKovynev:
strange why i don't see it in site? When run it in python server local all fine.

PR #2277, opened 2022-08-24
Update w2grid.js

peruginia:
added the style prameter to the columngroups (background color, text color...)

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -55,7 +55,7 @@ class w2grid extends w2base {
         this.name         = null
         this.box          = null // HTML element that hold this element
         this.columns      = [] // { field, text, size, attr, render, hidden, gridMinWidth, editable }
-        this.columnGroups = [] // { span: int, text: 'string', main: true/false }
+        this.columnGroups = [] // { span: int, text: 'string', main: true/false, style:'background-color:#xxxxxx' }
         this.records      = [] // { recid: int(required), field1: 'value1', ... fieldN: 'valueN', style: 'string',  changes: object }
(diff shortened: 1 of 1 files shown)


src/w2grid.js
```diff
@@ -7061,6 +7061,9 @@ class w2grid extends w2base {
                     console.log('NOTICE: grid column.caption property is deprecated, please use column.text. Column ->', col)
                     col.text = col.caption
                 }
+				colg.applystyle = '';
```
vitmalina, line 7064: please use space for indentation

src/w2grid.js
```diff
@@ -7098,7 +7101,7 @@ class w2grid extends w2base {
                     let gText = w2utils.lang(typeof colg.text == 'function' ? colg.text(colg) : colg.text)
                     tmpf      = '<td id="grid_'+ obj.name + '_column_' + ii +'" class="w2ui-head" col="'+ ii + '" '+
                            '        colspan="'+ colspan +'">'+
-                           '    <div class="w2ui-col-group">'+
+                           '    <div class="w2ui-col-group"' + colg.applystyle + '>'+
```
vitmalina, line 7104: I think it would be better to do use `colg.style` here directly, rather then assigning it to a variable above
`... style="' + colg.style + '">...`

src/w2grid.js
vitmalina: And these two lines should be removed

vitmalina (changes requested):

vitmalina:
Implemented this feature in a different PR

PR #2283, opened 2022-08-30
fix #2279

### LINKED ISSUE #2279: problem in format function in w2field.js
The online demo and if i don't change locale it works because group is equal at this.options.groupSymbol and the regExp is never called

i put a breakpoint in function format() at line 820 in w2field.js or at line 20970 in w2ui.js
   
in the console i have : 
   val = '20,000%'
   group = '.'
   this.options.groupSymbol = ','

the command :
      val.replace(new RegExp(group, 'g'), this.options.groupSymbol)

return
      ',,,,,,,' 

i have replaced with :
     val.replace('/' + group + '/g', this.options.groupSymbol)

and it return (corrected)
    '20,000%'

AlexKovynev:
fix https://github.com/vitmalina/w2ui/issues/2279

### DIFF
--- a/src/w2field.js
+++ b/src/w2field.js
@@ -819,7 +819,7 @@ class w2field extends w2base {
             // if default group symbol does not match - replase it
             let group = parseInt(1000).toLocaleString(w2utils.settings.locale, { useGrouping: true }).slice(1, 2)
             if (group !== this.options.groupSymbol) {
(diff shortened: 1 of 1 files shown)


AlexKovynev:
problem is "."(dot) cause group is not regexp. Replacing dot replace all chars.

vitmalina (approved):

Issue #2284, opened 2022-08-31
Blocking error in validating field with minLength


gregoriopellegrino:
**Short description**
When trying to validate a form with a field in which the `minLength` is set to a number and `required` option is set to `true`, if the field is blank an error is generated in the console and validation stucks.

**What is current behavior**
Validation fails. In the developer console I see:

```
Uncaught TypeError: Cannot read properties of null (reading 'length')
```

**What is desired behavior**
Validation corrected, reporting the error, indicating that the field requires a minimum number of characters.

**Link to jsfidle/codepan with sample code**
https://jsfiddle.net/r8f07Ln5/

To reproduce click on `Validate`

vitmalina:
This has been fixed. I just verified, it is working fine now

PR #2287, opened 2022-09-01
Fixing issue #2284

gregoriopellegrino:
https://github.com/vitmalina/w2ui/issues/2284

### DIFF
--- a/src/w2form.js
+++ b/src/w2form.js
@@ -755,6 +755,7 @@ class w2form extends w2base {
             }
             if (field.options && field.hidden !== true && field.options.minLength > 0
                     && ['enum', 'list', 'combo'].indexOf(field.type) == -1 // since minLength is used there too
+                    && val != null
                     && this.getValue(field.field).length < field.options.minLength) {
                 errors.push({ field: field, error: w2utils.lang('Field should be at least ${count} characters.', {count: field.options.minLength}) })
             }


AlexKovynev:
it seems it wrong fix cause it need to be show an error on null val too cause it will a zero characters.

gregoriopellegrino:
I'm not sure. We have two cases:

- field required and with minLength: in this case if the field is blank the user gets an error because the field is required
- field not required and with minLength: is it an error if the field is blank?

AlexKovynev:
Maybe you are right too. So will be @vitmalina decision:)

AlexKovynev:
@gregoriopellegrino it seems need to add val != '' too!

gregoriopellegrino:
The bug was in the control for `minLength`. The problem occurred when `val = null`, since `this.getValue(field.field).length < field.options.minLength` returned error, as the `null` value has no `length` property.

If `val = ''`, then checking t`his.getValue(field.field).length < field.options.minLength` works correctly, no changes needed.

I am not sure about your solution with `else if`, it would be to test in case a field has `minLength`, but is not `required`.

AlexKovynev:
@gregoriopellegrino Problem that if val == '' we have two errors but if val == null we only have first error with your code. With "else if" we have first error in both cases.

AlexKovynev:
That why i suggest https://github.com/vitmalina/w2ui/pull/2296.
Cause with your code if val == '' we still have both errors, but not with val == null.

AlexKovynev:
Maybe i better agree with you. BUT i rewrite my PR. Do you see the problem? 
.Null equal to ''. They both blank. So we need to ignore they both. But why? That why i vote for https://github.com/vitmalina/w2ui/pull/2296

vitmalina:
Guys, sorry, it took me some time to review. I am closing two other PRs and let finish up this here. Thank you for considering all possible case of this issue. Here is what I think:

We need to display min length error, if
- field is not hidden
- field is not ['enum', 'list', 'combo']
- field.options.minLengh is defined and over 0
- value is null or length is less then specified

Need to consider if value is
- a number
- a boolean

I think the condition should be like this

```js
if (field.hidden !== true && field.options?.minLength > 0
   && !['enum', 'list', 'combo'].includes(field.type) // since minLength is used there for other purposes
   && (val == null || String(val).length < field.options.minLength)) {
```
If you agree, can you please do these modifications, and I will merge the change.

AlexKovynev:
As for me i am completly agreed. It was one of mine PR. So i create new. You can merge it and close this [fix](https://github.com/vitmalina/w2ui/pull/2299)

vitmalina:
merged a fix with different PR

PR #2298, opened 2022-09-11
Fixing issue #2284 alt 2

### LINKED ISSUE #2284: Blocking error in validating field with minLength

AlexKovynev:
Fix https://github.com/vitmalina/w2ui/issues/2284

### DIFF
--- a/src/w2form.js
+++ b/src/w2form.js
@@ -755,6 +755,7 @@ class w2form extends w2base {
             }
             if (field.options && field.hidden !== true && field.options.minLength > 0
                     && ['enum', 'list', 'combo'].indexOf(field.type) == -1 // since minLength is used there too
+                    && val != null && val != ''
                     && this.getValue(field.field).length < field.options.minLength) {
                 errors.push({ field: field, error: w2utils.lang('Field should be at least ${count} characters.', {count: field.options.minLength}) })
             }


src/w2form.js
```diff
@@ -755,6 +755,7 @@ class w2form extends w2base {
             }
             if (field.options && field.hidden !== true && field.options.minLength > 0
                     && ['enum', 'list', 'combo'].indexOf(field.type) == -1 // since minLength is used there too
+                    && val != null && val != ''
```
vitmalina, line 758: Well, since val is already calcu;ated from `getField(field.field)`, I think we could just do this `&& val != null && val.length < field.options.minLength` and it should cover all casses

vitmalina:
Thank you for providing multiple solutions, see my comments here: https://github.com/vitmalina/w2ui/pull/2287

PR #2315, opened 2022-10-10
Draft: Allow strings and HTMLNodes to be inserted into the DOM

Kreijstal:
I tried to use https://w2ui.com/web/demos/#/toolbar/9  and the problem was that {type:"html"} doesn't accept a Dom Node


Kreijstal:
Now that I notice does this mean I have to use gulp?

Kreijstal:
okay so these are 2 commits the first one is the actual change I did, the second one is the change + gulp

vitmalina:
I have doubts that this change will fix the issue. I looked in the toolbar code and it does generate html and then inserts all into the dom. If you want to insert a DOM node into toolbar, you need to do it in `onRefresh` event. I have updated demo page to show how to do it, see, https://w2ui.com/web/demos/#/toolbar/9

PR #2332, opened 2022-11-09
Fix some bugs and complete todo item (and more)

ryanjbudhu:
While updating a project that used w2ui v1.3 I found a few bugs that were easy to fix.

For the grid toolbar I added support for reordering the default buttons by setting the order via `options.toolbar.items`. Take a look and see if you think it's worth adding (I can update any docs to describe what my comments say if need be for clarity).

I also extended the `w2utils.getSize()` function to be able to query any attribute (else return 0) since just getting the "width" attribute was causing w2objects within layouts to be too big (which either added scrollbars or clipped the content, depending on my overflow value).

One last change was to the liveSearch feature in grids. I just added a debounce function to reduce the amount of API calls when typing, works as expected!

If you have any questions/suggestions let me know, my team and I love this library and we thank you for keeping it updated!


src/w2field.js
```diff
@@ -492,7 +494,7 @@ class w2field extends w2base {
                 }, 1)
             }
             // if readonly or disabled
-            if (query(this.el).prop('readonly') || query(this.el).prop('disabled')) {
+            if (query(this.el).prop('readOnly') || query(this.el).prop('disabled')) {
```
vitmalina, line 497: Strictly, it is not necessary because of this: https://github.com/vitmalina/mquery/blob/main/src/mquery.js#L83. JQuery also does it.

src/w2field.js
```diff
@@ -1613,15 +1615,14 @@ class w2field extends w2base {
                     <input name="attachment" class="file-input" type="file" tabindex="-1"'
                         style="width: 100%; height: 100%; opacity: 0" title=""
                         ${this.options.max !== 1 ? 'multiple' : ''}
-                        ${query(this.el).prop('readonly') ? 'readonly': ''}
-                        ${query(this.el).prop('disabled') ? 'disabled': ''}
+                        ${query(this.el).prop('readOnly') || query(this.el).prop('disabled') ? 'disabled': ''}
```
vitmalina, line 1618: not sure why this is needed, but I think it would not do any harm. I rarely use file upload fields myself.

src/w2field.js
```diff
@@ -1732,7 +1733,7 @@ class w2field extends w2base {
         if (edata.isCancelled === true) return
         // if errors and not silent
         if (options.silent !== true && errors.length > 0) {
-            w2tooltip.show(this.el, 'Errors: ' + errors.join('<br>'))
+            w2tooltip.show(this.el, {html: 'Errors: ' + errors.join('<br>')})
```
vitmalina, line 1736: I think it should be 
```js
w2tooltip.show({
    anchor: this.el,
    html: ...
}
```

vitmalina: I will check and fix in another PR

src/w2grid.js
```diff
@@ -6259,6 +6262,11 @@ class w2grid extends w2base {
                 }
                 this.toolbar.items.push(w2utils.extend({}, this.buttons.save))
             }
+            // fill in overwritten items with default buttons
+            // ids are w2ui-* but in this.buttons the map is just [add, edit, delete]
+            // must specify at least {id, name} in this.toolbar.items if you want to keep order
+            tb_items = tb_items.map(item => this.buttons[item.name]
+                                            ? w2utils.extend({}, this.buttons[item.name], item) : item)
```
vitmalina, line 6269: Sounds good

src/w2layout.js
```diff
@@ -771,8 +771,14 @@ class w2layout extends w2base {
...
 
         // layout itself
+        // width includes border and padding, we need to exclude that so panels
+        // are sized correctly
         let width  = w2utils.getSize(query(this.box), 'width')
+                    - ( w2utils.getSize(query(this.box), 'borderRight') + w2utils.getSize(query(this.box), 'borderLeft') +
+                        w2utils.getSize(query(this.box), 'paddingRight') + w2utils.getSize(query(this.box), 'paddingLeft') )
         let height = w2utils.getSize(query(this.box), 'height')
+                    - ( w2utils.getSize(query(this.box), 'borderTop') + w2utils.getSize(query(this.box), 'borderBottom') +
+                        w2utils.getSize(query(this.box), 'paddingTop') + w2utils.getSize(query(this.box), 'paddingBottom') )
```
vitmalina, line 781: Lets do this fix like in this commit: https://github.com/vitmalina/w2ui/commit/0114843e051f75760bc324c71b5289a08c8f4c14

src/w2utils.js
```diff
@@ -2219,6 +2222,19 @@ class Utils {
             })
         })
     }
+
+    debounce(func, wait) {
```
vitmalina, line 2226: I like the idea of `debounce` here, it makes sense to have it in w2utils. Can we use this implementation of `debounce`, https://www.freecodecamp.org/news/javascript-debounce-example/.

I checked it keeps the right `this`, it has arrow functions and it is clean

src/w2utils.js
```diff
@@ -2219,6 +2222,19 @@ class Utils {
...
+        let timeout
+        return function executedFunction(...args) {
+            const later = () => {
+                clearTimeout(timeout)
```
vitmalina, line 2230: I think this clearTimeout is not necessary

vitmalina (changes requested):

vitmalina:
Thank you for the commit. I am glad w2ui works good for your team and you like it. You have many good finds in this commit, and I am glad it comes from a real conversion to v2 in a real project. Your fixes are much appreciated. I provided feedback and in some cases alternative fixes. Can you please rebase on master branch, review changes and submit it again.

I would be happy to talk more about your project and what you are doing with w2ui, find me on telegram @vitmalina. Thank you.

src/w2field.js
ryanjbudhu: I had just noticed that when I disabled file upload fields, the readonly field didn't stop functionality.

src/w2field.js
ryanjbudhu: Ah I thought it wasn't picking it up, good to know!

ryanjbudhu:
Thanks for reviewing! I added a change to the form.css as a candidate for the broken lock display as well, let me know if that works.

vitmalina (approved):
Looks good. Thank you

PR #2349, opened 2022-12-13
clean key down event when popup was closed

atolucky6:


vitmalina:
In order to submit PRs, you should describe what are you trying to do. Usually within an issue.

PR #2361, opened 2022-12-24
Bug fix #2350

### LINKED ISSUE #2350: w2form v2.0 - Generation of the change event in the absence of a change in the field value by the user, when a user enters a value in a field, attributes value.current and value.previous have the same value
w2form v2.0 - generates a change event for 
all of its fields when displayed in the 2layout panel 
or w2popup window (perhaps in the refresh() function).
The event object has the following values
```
event.detail.value.current: ""
event.detail.value.previous: undefined
```
or 
```
event.detail.value.current: ""
event.detail.value.previous: null
```
When generating the Change event on user input in the field, attributes event.detail.value.current and event.detail.value.previous have the same value, which is also an error.

heady-yar:
Fix generation of the change event in the absence of a change in the field value by the user

### DIFF
--- a/src/w2form.js
+++ b/src/w2form.js
@@ -1600,10 +1600,12 @@ class w2form extends w2base {
                         let helper = field.el._w2field?.helpers?.multi
                         query(helper).removeClass('w2ui-error')
                     }
-                    if (this._previous != null) {
+                    if (typeof(this._previous) != 'undefined') {
                         value.previous = this._previous
                         delete this._previous
                     }
+                    else
+                        return
                     // event before
                     let edata2 = self.trigger('change', { target: this.name, field: this.name, value, originalEvent: event })
                     if (edata2.isCancelled === true) return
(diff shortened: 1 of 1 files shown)


vitmalina:
I do not see where this change would actually make difference. Can you provide a use case with where the original code is not working before the changed and working after the change. Link to jsFiddle would suffice.

heady-yar:
I had a form on w2ui v1.5 in which the Save button became active only after changing the value of some field by the user. When transferring this form to w2ui v2.0, I found that additional Change events occur that are not related to user input.
Your suggestion is to remove line 1874 from w2form.js gives an excellent result - I have already checked it. Thanks.

vitmalina:
I thought more about it and I think it would be right things to do. Thanks for submitting it as a bug and fixing it. Can you please submit PR with just that change and I will merge it. Thank you

heady-yar:
Okay, I'll do it.

Issue #2320, opened 2022-10-14
Custom Fields - prefix and suffix not clickable 


foxpcteam:
Hi,
when you use the "Custom Fields" with prefix and suffix those elements become not clickable, and you can't do any actions based on click.

The bug can be seen in the demo page:
https://w2ui.com/web/demos/#/fields/9

Try to click on the "go" button and see what node you will get.
You can check this with click listener:

`window.addEventListener('click', (e) => console.log(e.target));
`

blitz-research:
Hi,

This can be fixed by enabling mouse events in the button's 'style', eg: ...style="pointer-events: all;...

I've made a PR which adds prefix/suffix styles to w2field that does this as it seems like it'd be the preferred default behaviour, not sure how 'safe' it is though, I have little experience with this stuff.

vitmalina:
I still would prefer that the default prefix and suffix would not have any pointer events, unless it is needed. I have fixed the example mentioned in the issue to make button work.

PR #2368, opened 2022-12-28
Added w2field prefix/suffix styles

blitz-research:
To address #2320, added w2field prefix and suffix styles which enable mouse events.


vitmalina:
Thank you for contribution. However, I prefer that the default prefix and suffix would not have any pointer events, unless it is needed. I have fixed the example mentioned in the issue to make button work.

PR #2370, opened 2023-01-06
we should define var before usage

alexkuznetsov:
Hi, minor fix.
Problem happens when grid selection mode is configured as `cell` mode and we try to  open context menu.

image

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -4366,6 +4366,7 @@ class w2grid extends w2base {
         if (this.selectType == 'row') {
             if (sel.indexOf(recid) == -1) this.click(recid)
         } else {
+            let selected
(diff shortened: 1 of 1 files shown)


AlexKovynev:
let selected = false ?

alexkuznetsov:
> let selected = false ?

Well, because of `!selected`, i thought it should be ok without initial assignment.

Could this be a potential problem? We tested it, no issues so far...

AlexKovynev:
Maybe cause codestyle:) In other places "let selected = false" used by Vitaliy. Anyway i create another PR. He can choose.

vitmalina (approved):

vitmalina:
Already merged

PR #2373, opened 2023-01-25
we neet to translate button labels before use them in this.confirm

alexkuznetsov:
Otherwise they will be as they are, without translation


vitmalina (approved):

alexkuznetsov:
Thanks!

PR #2380, opened 2023-02-10
When search field is list, we should display text

alexkuznetsov:
Hi there!
When the search field is a list, the text should be displayed, not the value (unless they match)

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -5076,6 +5076,12 @@ class w2grid extends w2base {
                 let display
                 if (Array.isArray(sd.value)) {
                     display = `<span class="grid-search-count">${sd.value.length}</span>`
+                } else if (sf && sf.type == 'list') {
+                    const sd_value = sd.value
+                    const sd_text = sd.text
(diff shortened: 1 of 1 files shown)


src/w2grid.js
```diff
@@ -5076,6 +5076,12 @@ class w2grid extends w2base {
                 let display
                 if (Array.isArray(sd.value)) {
                     display = `<span class="grid-search-count">${sd.value.length}</span>`
+                } else if (sf && sf.type == 'list') {
+                    const sd_value = sd.value
```
vitmalina, line 5080: I do not see a point to define constants here, why not 
```js
display = !!sd.text && sd.text !== sd.value ? `: ${sd.text}` : `: ${sd.value}`
```

alexkuznetsov: Well, you absolutely right, updated PR here.

vitmalina (approved):

alexkuznetsov:
@vitmalina thanks!

Issue #2381, opened 2023-02-10
[BUG] w2ui grid: load more spinner not resized to full width - NOTE: easy to fix!


H-Lo:
**Short description**
Load more spinner is not resized to full width.

w2ui grid version: **2.0.nightly.2023-01-16**

**Steps to reproduce or sample** 
Open any grid that has to load more records and see a spinner.

<hr/>

The problem is "display:block" is used for table row TR to show the spinner:

```
<tr id="grid_GRID_NAME_rec_more" style="display: block;">
	<td colspan="2000" class="w2ui-load-more">
		<div>
			<div style="width: 20px; height: 20px;" class="w2ui-spinner"></div>
		</div>
	</td>
</tr>
```

The right way to show it would be "<b>display:table-row;</b>"

The bug can be fixed by adding this CSS line:

`tr#grid_GRID_NAME_rec_more { display:table-row !important; }`

But it would be better to have it fixed rather than fixing it outside w2ui grid especially because **!important** modifier is not a good way to go.

vitmalina:
PR request with a fix merged

PR #2382, opened 2023-02-13
Fixing spinner load more

H-Lo:
Fixing issue:
https://github.com/vitmalina/w2ui/issues/2381

### DIFF
--- a/src/query.js
+++ b/src/query.js
@@ -519,12 +519,12 @@ class Query {
     show() {
         return this.toggle(true)
     }
-	showTr() {
-		return this.each(node => {
-			node.style.display = 'table-row'
-		})
-		return this
-	}
+    showTr() {
+        return this.each(node => {
+            node.style.display = 'table-row'
+        })
+        return this
+    }
     hide() {
         return this.toggle(false)
     }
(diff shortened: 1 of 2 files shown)


src/query.js
```diff
@@ -519,6 +519,12 @@ class Query {
     show() {
         return this.toggle(true)
     }
+    showTr() {
```
vitmalina, line 522: query here is taken from another project, https://github.com/vitmalina/mquery. It should not be modified just here. It should also be modified in the original project. Also, I do not think adding `showTr` is a good idea. It should be still within `show` possibly check if element is a TR, then do display: 'table-row'

vitmalina (changes requested):

src/query.js
H-Lo: Well yes, everything you said is right. But so far I don't know enough about mQuery to fix it. Tried to look into but not that I'm expert on JavaScript. Maybe I can take a loog again in few days and see is there a chance for me to fix it.

H-Lo:
@vitmalina
Ok, I think I did it right this time:
1) fixed `mquery.js` in `mquery` project to handle TR diferrently than other elements
2) reverted everything about `showTr()` in `w2ui` project
3) made changes to `query.js` inside `w2ui` project to be the same as the on in `mquery` project

vitmalina (approved):

PR #2383, opened 2023-02-13
Fix inline editing for 'list' types in grid (fixes #2378).

### LINKED ISSUE #2378: Grid inline editing type : list in v2.0
**Short description**
In 2.0 : input types "list", "enum" and "file" are not supported in inline grid editing."

**What is current behavior**
{ field: 'ageunit', text: 'Age Unit', size: '75px', editable: { type: 'list', items: [{ id: 'Day', text: 'Day' },{ id: 'Month', text: 'Month' }] } },
It returns 'undefined'

**What is desired behaviour**
What I have to use now instead "List" type? Combo do not fit, because can't returns selected item id. Or, can you give me the way, please, to get id of selected item in combo list?

**Link to jsfidle/codepan with sample code**
...

**Steps to reproduce or sample** 

...

ryanjbudhu:
Started using inline editing when I realized 'list' was unsupported. Some styling also needed to be changed to display it correctly, I've tested to make sure it doesn't break anything but let me know if there was a reason.

I also updated the grid demo (tab 21) to show off both combo and list.

### DIFF
--- a/demos/examples/grid/21.html
+++ b/demos/examples/grid/21.html
@@ -10,7 +10,7 @@
 </div>
 
 <!--CODE-->
-<div id="grid" style="width: 900px; height: 400px;"></div>
+<div id="grid" style="width: 1000px; height: 400px;"></div>
 <br>
 <label style="margin-left: 20px">
     <input type="checkbox" id="fixedsize" checked="checked" onclick="advanceOnEdit(this.checked)">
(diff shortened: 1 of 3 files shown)


vitmalina:
Well, never had enough time to finish implemented it. I pulled your branch, tested a bit. It is a good start, I noticed a couple of things that I wished worked.
1. When I try type to filter in the list, I do not see text I type or the cursor. It would be nice to be see it
2. When I selected something, I think it should advance to the next row and select input, just like other fields do that
3. When it is in focus, it does not show the blue border around it
4. I should be able to hit tab (or tab+shift) and go from editable field to editable field. However, if there is a list, it breaks it.

But definitely a good start.

ryanjbudhu:
Thanks! For at least the first one, the change to the CSS should've fixed that but I'll take another look.
As for the rest I noticed the field is a bit fiddly since focus seems to be pretty sensitive so I can take a closer look at that.

PR #2393, opened 2023-03-16
w2grid enhancement - cache query(this.box)

Yahasana:


vitmalina:
Thank you for your PR. I agree that the syntax with `$box` is more expressive then always write `query(this.box)`, however I am reluncant to accept this PR because of two reasons:
1. I do not think it will do any reasonable speed improvements because in majority of cases `this.box` is just an html element and `query(this.box)` returns immediate result without DOM parsing. Plus, I did not see argumentation why you changed it.
2. The change is too involved and we do not have very comprehensive test coverage and we are close to 2.0 release, therefore, I think it is not the best time

Yahasana:
> 1. I do not think it will do any reasonable speed improvements because in majority of cases `this.box` is just an html element and `query(this.box)` returns immediate result without DOM parsing. Plus, I did not see argumentation why you changed it.

`query()`  create a new object each time. see [L564](https://github.com/vitmalina/w2ui/blob/master/src/query.js#L564)

vitmalina:
Yes, you are right. I will create a new `query` object, but if you look here https://github.com/vitmalina/w2ui/blob/master/src/query.js#L10, you'll see that it will just create a new object with a single node in it w/o DOM traversing. 

Do you have some estimates or a use case where it could be measured to show that this will provide noticeable speed improvements?

vitmalina:
I am a bit reluctant to merge it now, as we are close to 2.0 release and this change has large code impact and will require more through testing.

vitmalina:
Closing due to no response

Issue #2399, opened 2023-04-24
W2Grid: Error when loading grid while cursor is over column header at render


TakoOu:
**Short description**
When loading the grid while the cursor is over a column header (with a tooltip), sometimes it generates an error as "item" is undefined.

**What is current behavior**
Expected behavior with no visual issue, but it generates an error in the browser console

**What is desired behavior**
Avoid generating an error

**Link to jsfidle/codepan with sample code**
I was not able to reproduce with a simple fiddle, I believe it might needs an heavier load (datawise and codewise as I am rerouting request event in my case)

**Steps to reproduce or sample** 
- Loading the grid from an interface element over the column header row, as per this screenshot:
![image](https://user-images.githubusercontent.com/24452494/233961908-1aec82d2-f226-4a79-bb3f-275b7ac2c8a5.png)
- Sometimes this will generate an error in the browser as "item" is undefined while tooltip is begin created:
![image](https://user-images.githubusercontent.com/24452494/233962751-90cbf203-c0d0-427f-ae4a-baed25340ef5.png)

I was able to avoid the error by simply adding an optionnal chaining to the tooltip reference. (see incoming PR)

vitmalina:
Fixed merged

PR #2400, opened 2023-04-24
W2Grid: Avoid an error when the grid is loaded while cursor is over a column as it is rendered

TakoOu:
Avoid an error when the grid is loaded while cursor is over a column as it is rendered

Proposed "fix" for issue #2399


vitmalina (approved):

vitmalina:

PR #2401, opened 2023-04-24
W2Grid: Added possibility to interact with the contextmenu on column header event.

TakoOu:
Added possibility to interact with the contextmenu on column header event.

For some context:
I was using the contextmenu event in 1.5 to show a specific menu when right clicking on column header.
In the current state of 2.0 this was not possible anymore.

I tested it by adding a handler through the onColumnContextMenu property as well as with the on('columnContextMenu', ...) method. Works fine in both case.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -331,55 +331,56 @@ class w2grid extends w2base {
         }
 
         // events
-        this.onAdd              = null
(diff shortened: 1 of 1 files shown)


vitmalina:
Looks good. I would only suggest to add documentation for the new method and event.

PR #2423, opened 2023-07-06
Update w2grid: sort method. handle unsort after sort desc

luis40895:
Use case : remove a column sort.
Like this the sort is : asc -> desc -> no sort -> asc -> desc -> no sort -> ...
Today :  asc -> desc ->  asc -> desc -> ...

Works with multifield.

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -4633,6 +4633,7 @@ class w2grid extends w2base {
         // check if needed to quit
         if (field != null) {
             // default action
+            var removeSort = false;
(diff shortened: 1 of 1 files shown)


src/w2grid.js
```diff
@@ -4633,6 +4633,7 @@ class w2grid extends w2base {
         // check if needed to quit
         if (field != null) {
             // default action
+            var removeSort = false;
```
vitmalina, line 4636: you should use `let` in the future please

vitmalina (approved):

PR #2432, opened 2023-08-19
w2grid load data fail

WenSheng-tw:
When loading data from the backend fails, the console will show an error message!!

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -2911,6 +2911,7 @@ class w2grid extends w2base {
     }
 
     requestComplete(data, action, callBack, resolve, reject) {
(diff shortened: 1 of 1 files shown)


vitmalina:
From the description of the PR, it is not clear what you tried to achieve. If you wanted to suppress console messages, you should fix your backend instead. I found those messages useful.

alexkuznetsov:
In my opinion, a silent return, as in this PR, can bring more problems.
The Grid API requires a specific schema for data from the backend API, so if you want to receive data from any backend, that backend must meet those requirements.

vitmalina:
Closing this PR due to no response from the author

Issue #2508, opened 2024-03-25
w2ui grid - password renderer error if password undefined


ope-nz:
Hello, when inserting a new row into the grid (without a template) the password renderer is throwing an error as the value is "undefined"

**What is current behavior**
The error below is thrown if the value is undefined and the new row is not added.

![image](https://github.com/vitmalina/w2ui/assets/26259049/0a88340f-bc73-46ef-b34d-f6a52d01ed14)

**What is desired behavior**
The password renderer should be blank until the password value is set.

**Link to jsfidle/codepen with sample code**
https://jsfiddle.net/xw0Lznch/

**Steps to reproduce or sample** 
See console output in codepen

**Proposed fix**

In w2utils.js update 'password' renderer to;

'password'(value, params) {
                let ret = ''
                if (!value) return ret;
                for (let i = 0; i < value.length; i++) {
                    ret += '*'
                }
                return ret
            }

ope-nz:
I have created a pull request

https://github.com/vitmalina/w2ui/pull/2509

ope-nz:
I have created a pull request

https://github.com/vitmalina/w2ui/pull/2509

PR #2509, opened 2024-03-25
Update w2utils.js

ope-nz:
Fix password renderer for new rows when cell value is undefined


ope-nz:
Relates to issue https://github.com/vitmalina/w2ui/issues/2508

PR #2517, opened 2024-05-11
BUGFIX Violation (non-passive event listener)

H-Lo:
Fixed but that was resulting as console log warning while initializing grid element.

```
[Violation] Added non-passive event listener to a scroll-blocking 'mousewheel' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952
```

### DIFF
--- a/src/query.js
+++ b/src/query.js
@@ -381,6 +381,9 @@ class Query {
             }
             this.each(node => {
                 this._save(node, 'events', [{ event, scope, callback, options }])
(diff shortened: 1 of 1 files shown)


H-Lo:
violation

vitmalina:
Closing as no response

H-Lo:
Hi @vitmalina!
"Closed as no response"  - which response was expected? I did not see any questions.

Issue #1417, opened 2016-12-12
Scrolling with mouse wheel does not work on frozen columns


mpf82:
If you scroll up/down with the mouse wheel while the cursor is over a frozen column, the grid will not scroll.

http://jsfiddle.net/9w6LLbup/

I'm pretty sure it did work in the past ...

aswzen:
so how its going? any update?

PR #2524, opened 2024-05-21
Fix scrolling on frozen column with mouse wheel on firefox

mpf82:
Releated issue: https://github.com/vitmalina/w2ui/issues/1417

Reasoning for this change:
"typeof" always returns a string, so the check always evaluated to true, thus using "event.wheelDelta" on firefox, which is undefined.


vitmalina:
Good to see you again!

PR #2555, opened 2024-08-30
#2554 fix and refactor date tooltip getMonthHTML

### LINKED ISSUE #2554: Weekdays displayed incorrectly in the date picker widget
**Short description**
There is an invalid weekday offset in the form date picker widget.
I use Monday as a first day of the week, and noticed this for September 2024.
1st of September should be on Sunday, instead of Monday.

**Link to jsfiddle with sample code**
https://jsfiddle.net/qhsc1trv/

Screenshot 2024-08-30 at 00 01 32

image

dv1x3r:
I did a brief refactoring for `getMonthHTML` function. fixes #2554
1. The bug was located between w2tooltip.js lines 2760 and 2770
2. If `weekDay = 0` (Sunday), no offset was applied for Monday-first calendars
3. The `else` block at line 2765 does not make sense for me
4. I also changed some variable names to prevent confusion:
- Now we operate with only one `DT` object, keeping all original transformations
- `weekDays` -> `weekDaysHeaderHTML`
- `weekDay` and `weekday` -> `weekDayOffset`

(PR does not include recompiled files)

### DIFF
--- a/src/w2tooltip.js
+++ b/src/w2tooltip.js
@@ -2711,31 +2711,32 @@ class DateTooltip extends Tooltip {
             sdays.unshift(sdays.pop())
         }
 
-        let td = new Date()
+        let DT = new Date()
         let dayLengthMil = 1000 * 60 * 60 * 24
+
         let selected = options.type === 'datetime'
             ? w2utils.isDateTime(options.value, options.format, true)
             : w2utils.isDate(options.value, options.format, true)
         let selected_dsp = w2utils.formatDate(selected)
+
         // normalize date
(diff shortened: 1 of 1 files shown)


src/w2tooltip.js
```diff
@@ -2711,31 +2711,32 @@ class DateTooltip extends Tooltip {
             sdays.unshift(sdays.pop())
         }
 
-        let td = new Date()
+        let DT = new Date()
```
vitmalina, line 2714: `td` stands for `today`

vitmalina (approved):

PR #2588, opened 2024-11-18
Fix for incorrect url resolution and single quotes for a string 

alexkuznetsov:
In this PR:

 - Refactored url resolution for the grid;
 - Replaced double quotes with single quotes for `undefined` string.

The problem is with the current method used to resolve urls in the data grid.
AFAIK, in some cases url's are resolved more securely, but not all.

Minimal source code to understand the problem:

```js

class MinExample {

    constructor(o) {
        this.url = o.url;
        Object.assign(this, o?.settings ||| {})
    }

    refresh() {
        let url = (typeof this.url != ‘object’ ? this.url : this.url.get);
        console.log(‘url’, url);
    }

}

```

In the refresh method, we resolve the url and just output it to the console.
This is fine if the url is an object, string or whatever, but it breaks **if the value of this.url is null**. The error for a null value is as follows:

```
Uncaught TypeError: Cannot read properties of null (reading ‘get’)
    at x.refresh (<anonymous>:9:66)
    at <anonymous>:1:5
``` 

The fix also replaces double quotes for `undefined`:
```js
let amount = typeof event.wheelDelta != "undefined" ? -event.wheelDelta : (event.detail || event.deltaY)
``` 
becomes

```js
let amount = typeof event.wheelDelta != 'undefined' ? -event.wheelDelta : (event.detail || event.deltaY)

```

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -3677,7 +3677,7 @@ class w2grid extends w2base {
             return
         }
         // call delete script
-        let url = (typeof this.url != 'object' ? this.url : this.url.remove)
+        let url = this.url?.remove ?? this.url
         if (url) {
             this.request('delete')
         } else {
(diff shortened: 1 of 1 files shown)


alexkuznetsov:
@vitmalina should i create a new PR with rebuilded dist files?

PR #2591, opened 2024-12-01
For sidebar and toolbar, a cursor appears when you hover the mouse ov…

mansurdegirmenci:
…er the menus.

Added 'break' element to 'toolbar' in index.js. Changed unsupported substr method to slice.


demos/index.js
```diff
@@ -34,15 +34,25 @@ query(async function () {
             name: 'demo_toolbar',
             items: [
                 { id: 'combo', type: 'radio', text: 'Combo', icon: 'fa fa-star-o', route: 'combo/1' },
+                { type: 'break'},
```
vitmalina, line 37: break after each toolbar item does not look good (in my opinion). Break make sense to visually separate groups of similar buttons. I would prefer not to have break here.

demos/index.js
```diff
@@ -534,7 +544,7 @@ query(async function () {
     function hashChanged() {
         query('#drag-me').remove()
         // show toolbar
-        let hash = String(document.location.hash).substr(2)
+        let hash = String(document.location.hash).slice(2)
```
vitmalina, line 547: I am aware that `substr` supposedly deprecated, it was like that last 10 years

dist/w2ui.css
```diff
@@ -1,4 +1,4 @@
-/* w2ui 2.0.x (nightly) (10/31/2024, 10:55:25 AM) (c) http://w2ui.com, vitmalina@gmail.com */
+/* w2ui 2.0.x (nightly) (12/3/2024, 10:29:41 PM) (c) http://w2ui.com, vitmalina@gmail.com */
```
vitmalina, line 1: Please do not include files in `/dist` folder with the same PR, they need to be in the PR of their own. Or not at all. as I frequently compile them too

src/w2utils.js
```diff
@@ -1386,7 +1386,7 @@ class Utils {
      *      where: el/selector, // element or selector where to show, default is document.body
      *      timeout: int,       // timeout when to hide, if 0 - indefinite
      *      error: boolean,     // add error clases
-     *      class: string,      // additional class strings
+     *      side: string,       // right, left, center
```
vitmalina, line 1389: I think `class` is still useful. It should be kept

src/w2utils.js
```diff
@@ -1416,37 +1416,62 @@ class Utils {
...
+    * 
+    * - Resolves any pending Promise linked to the notification.
+    * - Clears the timeout associated with the notification, if one exists.
+    * - Removes the notification element (`#w2ui-notify`) from its parent container.
+    * - Resets all temporary references (`tmp`) to ensure proper cleanup.
+    * 
+    * This method ensures that any active notification is properly terminated, 
+    * preventing memory leaks or conflicts with future notifications.
+    */
+
```
vitmalina, line 1458: I do not like that this is part of `w2utils`, the only purpose of this is to clean up existing notify, which is also done if you call notify without parameters. Can we move it inside `notify` function and rename it to `_removeNofity` to indicate that it is internal. So, `w2utils.notify(msg)` displays a message and `w2utils.nofity()` hides it

vitmalina (changes requested):

vitmalina:
Thanks for you PR, see my comments

PR #2593, opened 2024-12-04
Fix parse int

alexkuznetsov:
Fixed a small typo in the function name.
Strings without interpolation should be singlequoted.

Also recompiled JS after all these changes.


vitmalina (approved):

vitmalina:
Thank you for your PR. In the future, please do not include files in `/dist` folder, as they need to be in a separate PR as they are generated

alexkuznetsov:
@vitmalina Sorry for that.

In my previous PR (#2588), I asked about this, but you didn't seem to get that comment. It's clear now, thanks!

PR #2597, opened 2024-12-11
notification various arrangements

mansurdegirmenci:
notification was removed from the w2utils class. shadowing was corrected. it was fixed to appear in div or body. new features were added. it was made open to development.


vitmalina:
Thank you for you contribution. But I think notification is small enough to be part of w2utils, not a stand alone class. I have been doing some update in it too. It should be working better. 

Thanks!

Issue #2599, opened 2024-12-17
Grid delete with force not passing "options" argument to "message" method


zbeebembg:
**What is current behavior**
Just recently I noticed an error being thrown when I try to delete records from a grid. After the delete confirmation, w2ui calls the grid delete method with "force" = true. This then calls the grid "message" method with no arguments (grid.message()). The message method then calls the utils message method, which takes two arguments "where" and "options". The call is failing within the utils message method, where it is checking logic for options.text and options.body. Below is the console traceback:

Uncaught TypeError: Cannot read properties of undefined (reading 'text')
    at Utils.message (w2ui.js:2039:22)
    at w2grid.message (w2ui.js:20144:24)
    at w2grid.delete (w2ui.js:15154:25)
    at w2ui.js:15174:32
    at w2base.<anonymous> (w2ui.js:2117:50)
    at w2base.trigger (w2ui.js:209:34)
    at options.action (w2ui.js:2235:33)
    at HTMLButtonElement.<anonymous> (w2ui.js:3245:49)

**What is desired behavior**
Maybe pass an empty or default options object when calling the grid message method?

**Link to jsfidle/codepan with sample code**
[W2UI Grid Delete Error - JS Fiddle](https://jsfiddle.net/zacharybeebe/r0h6xdyw/4/)

zbeebembg:
This has been fixed as of 01/03/2025, closing

PR #2600, opened 2024-12-18
bug fix for issue #2599

zbeebembg:
line 1111 of w2utils.js

Summary:
w2grid "message" method has an argument for "options" and that is passed to the Utils "message" method
however, sometimes the grid calls its "message" method without any arguments, therefore passing an
undefined argument as options to the Utils "message" method; the "arguments" keyword will still sees 2 arguments 
(arguments.length == 2) and previously wouldn't execute the logic of setting options = where, but with the extra logic
of checking if options === undefined, this should now execute as intended

### DIFF
--- a/src/w2utils.js
+++ b/src/w2utils.js
@@ -1108,7 +1108,12 @@ class Utils {
                 text  : String(options),
             }
         }
-        if (arguments.length == 1) {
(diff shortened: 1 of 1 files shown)


vitmalina:
This has already been fixed in the latest code. Nearly same fix

PR #2602, opened 2024-12-27
The button to reset the search must be of type="button".

alexkuznetsov:
Hello there!

The problem occurs when we wrap the grid with a form, and the buttons without type act as submits. 

Recompile of the JS is required, it is not included in the PR.


alexkuznetsov:
@vitmalina , is everything ok with this PR or does something need to be changed?

vitmalina (approved):

Issue #2617, opened 2025-03-19
Tab click throws error from missing "event" arg


zbeebembg:
**Short description**
I've just recently noticed this problem, first if I try to "click" a tab using the API, I need to pass in an empty object as the second argument after the tab id:  tabs.click(<id>, {}) otherwise it will throw an error.

And this problem persists internally, as well, for forms with tabs. It seems that when the form gets rendered it tries to click the first tab to activate it and again will throw an error because it is missing that second argument.

The tab click error is at line 10440 of the rawgit JS distribution, and the form error occurs when it calls the tab.click at line 22261.

**Link to jsfidle/codepan with sample code**
[JS Fiddle](https://jsfiddle.net/zacharybeebe/cayex37r/28/)

PR #2618, opened 2025-03-19
updated w2tabs.js line 574 - added a check for the "event" arg

zbeebembg:
This is in relation with Issue [#2617](https://github.com/vitmalina/w2ui/issues/2617)

### DIFF
--- a/src/w2tabs.js
+++ b/src/w2tabs.js
@@ -571,7 +571,7 @@ class w2tabs extends w2base {
     // -- Internal Event Handlers
 
     click(id, event) {
-        if (query(event.target).hasClass('w2ui-tab-close')) {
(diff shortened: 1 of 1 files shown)


src/w2tabs.js
```diff
@@ -571,7 +571,7 @@ class w2tabs extends w2base {
     // -- Internal Event Handlers
 
     click(id, event) {
-        if (query(event.target).hasClass('w2ui-tab-close')) {
+        if (event && query(event.target).hasClass('w2ui-tab-close')) {
```
vitmalina, line 574: More elegant solution might be `query(event?.target)...`

vitmalina (approved):

Issue #2622, opened 2025-04-01
grid add event


tbmale:
grid `add` event is not triggered by add method

PR #2623, opened 2025-04-01
trigger add event on add method in w2grid.js

tbmale:
... in relation with issue #2622

### DIFF
--- a/src/w2grid.js
+++ b/src/w2grid.js
@@ -498,6 +498,7 @@ class w2grid extends w2base {
         } else {
             this.refresh()
         }
+        this.trigger("add", {target: this.name})
         return added
     }
 


src/w2grid.js
```diff
@@ -498,6 +498,7 @@ class w2grid extends w2base {
         } else {
             this.refresh()
         }
+        this.trigger("add", {target: this.name})
```
vitmalina, line 501: I think it would be useful to pass on `record` and `first` variables with the event, as those are arguments in the functio. Also, add spacing after `{` and before `}`

Also, I noticed that there is already an add event (search the code) and it is when add button is clicked, so event name should be `addRecords`,

Also, you should have event's before and after loop. See how other events are triggered, with edata.finish(). Event should be cancellable too.

vitmalina (changes requested):

vitmalina:
Closing due to no response

PR #2639, opened 2025-06-13
Fix incorrect cleaning of numeric fields

medranocalvo:
The `groupSymbol` was being used literally in a regular expression, but in various locales it is the character `.` which is the character wildcard.

The result was that the field was immediately cleared upon blur.


vitmalina:
I fixed it in this commit https://github.com/vitmalina/w2ui/commit/6bfc0ff7e256a7e17fc0148aab035555eb48d8d5

Thanks for your contribution

PR #2644, opened 2025-08-31
Add golang server example

dv1x3r:
This PR introduces a new Go server example for w2ui with full CRUD support using in-memory SQLite database. It showcases how to integrate w2ui with Go through the external [w2go bindings](https://github.com/dv1x3r/w2go).

The following example is compact and self-contained yet demonstrates the full capabilities of w2ui remote data integration:
- w2grid + inline edit + row reorder
- w2form
- dropdowns :)


vitmalina:

